PR #13, opened 2020-11-14
Add HOME XY button and change in print.py for lower resolution

vacapetr:


jordanruthe (changes requested):
Home XY is missing the opening {

Can you update this and I will merge it?

PR #144, opened 2021-04-29
New Style

alfrix:
I have created a new icon-set with the solarized palette, and also changed the css to solarized, what do you think?

i may want bigger text so i'll take a look at that and the organization is a bit of a mess, we should put homing and movement together and move the custom temp and fans with preheat, i'll see what i can put  together in another pull

i left the z-bolt ready to have some switch to change themes


Elektrickser:
Hi.

Maybe you have Screenshots? 
How will you change the Icons? A Switch in the configuration or in the config file? 

The Symbols looks good.

jordanruthe:
Overall, I'm not a fan of any changes to the python files that will break the current theme unless it supports theming going forward. I believe there are a few things we need to do before merging this. 

* We need to have some standardization between themes for the image names. That way doing a new theme won't change python files. This might result in having to copy an image to multiple file names, but it will help with having multiple themes.
* We shouldn't need to change css classes between themes as the css file would account for this 
* Need a configuration option for theme. Ideally in the settings panel, but at least in the configuration file. 
* Might need a theme settings json file. Specifically, I'm thinking this for the title_spacing variable, but there may be other cases. It's possible the settings config file could map image name to file name so the same image could be referenced multiple times.

alfrix:
>Maybe you have Screenshots?

sure look:

tes
yes i have noticed the 1 is not centered on the image, but it is centered on the svg, i will try to workaround this later.
and no i don't have 5 extruders i was just testing

zcalibrate
preheat
Homing
config
actions


> _We need to have some standardization between themes for the image names._

agreed

>_That way doing a new theme won't change python files_ 

i had to modify the py files because of new icons and modified the icon selections where they didn't have much sense, i tried to use coherent naming.

> _This might result in having to copy an image to multiple file names_

 please no. if an icon is missing we should draw an ? or a default icon and not have duplicates.

> _Need a configuration option for theme. Ideally in the settings panel, but at least in the configuration file._

Ok

jordanruthe:
> > This might result in having to copy an image to multiple file names
> please no. if an icon is missing we should draw an ? or a default icon and not have duplicates.

I'm not trying to have duplicate image files, but in the case of the current z-bolt theme, 'decrease' is used for temperature decrease and also 'cancel'. I'm saying we need to have standard names, so as a temporary patch the cancel and decrease images may be the same for now, but they should be discrete files. This will allow themes to have the right image for the right case.

alfrix:
> the current z-bolt theme, 'decrease' is used for temperature decrease and also 'cancel'.
yes, i have seen it and i'm creating new z-bolt icons and updating it to be more consistent
i have added quite a few of new icons that were needed, if i missed one please tell i like to draw :)

to avoid duplicates or wrong images: 
if file !exist:
 log "missing file"
skip image and draw label only

alfrix:
Z-bolt updated in this PR: (it has it's own css) maybe a few more tweaks are needed
bed
extrude
macros
main
menu
move
offset
system

alfrix:
For me this its ready, new icons, with the ability to swap theme (icons and css) 
do you want me to drop the title_spacing change?

i want to make some progress on the ui scale but on another PR and after some experiments

jordanruthe:
In the z-bolt theme, I'd like to keep the home icon on the left the house still. 
Let's also drop the title spacing change for now, and I want to keep z-bolt as the default theme.

Otherwise, I think this is ready to merge

screen.py
```diff
@@ -109,7 +109,8 @@ def __init__(self, args, version):
         self.set_resizable(False)
         logging.info("Screen resolution: %sx%s" % (self.width, self.height))
 
-        self.gtk = KlippyGtk(self, self.width, self.height)
+        self.theme = self._config.get_main_config_option('theme')
```
jordanruthe, line 112: Can you change this to: `self._config.get_main_config_option('theme','z-bolt')`

PR #150, opened 2021-05-10
M117 messages display

zztopper:
Added displaying M117 messages under printer state label on JobStatus panel.

### DIFF
--- a/panels/job_status.py
+++ b/panels/job_status.py
@@ -381,9 +381,7 @@ class JobStatusPanel(ScreenPanel):
 
         ps = self._printer.get_stat("print_stats")
         vsd = self._printer.get_stat("virtual_sdcard")
-        msg = self._printer.get_stat("display_status")
-        if "message" in msg:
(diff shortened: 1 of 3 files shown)


panels/job_status.py
```diff
@@ -588,6 +595,9 @@ def update_text(self, label, text):
 
     def update_progress (self):
         self.labels['progress_text'].set_text("%s%%" % (str( min(int(self.progress*100),100) )))
+    def update_message (self):
+        msg = self._printer.get_stat("display_status")
+        self.labels['lcdmessage'].set_text("%s" % (msg['message']))
```
jordanruthe, line 600: You may as well use get_stat such as get_stat("display_status","message")

jordanruthe: Also, instead of `"%s" % (msg['message'])` I would suggest something like `"" if msg == None else msg` so that if there is no message, it doesn't display anything

PR #151, opened 2021-05-10
Display speed

zztopper:
Removed current G-code move X and Y coordinates, moved Z coordinate to the left, added current G-code speed in mm/s to the right.

JobStatus panel now looks like this:
telegram-cloud-photo-size-2-5247246783900398503-y

### DIFF
--- a/ks_includes/printer.py
+++ b/ks_includes/printer.py
@@ -76,6 +76,7 @@ class Printer:
     def process_update(self, data):
         keys = [
             'bed_mesh',
+            'display_status',
             'fan',
             'gcode_move',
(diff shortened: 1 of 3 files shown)


nicedevil007:
Will this update as you increase the speed under finetuning?

Also may I ask if it would be possible to show the actual layer instead of Z-height?

hapklaar:
Sweet! Was just looking for the option to show current speed on KlipperScreen. But how should I configure it so it does, as my latest version of KlipperScreen still shows XYZ pos.
Are there configuration options I should use (and am not aware of)?

Issue #157, opened 2021-05-16
Autoreload KlipperScreen after a theme change was chosen


nicedevil007:
heya @alfrix,

maybe it is just for my installation but I guess not. After changing to your new addition to KipperScreen I had to restart KlipperScreen to get it working.

Maybe that can be done automatically if anyone will change this setting?

alfrix:
maybe a dialog saying "Restart required, do you want to restart now?" or maybe just restart it's only the gui, it shouldn't affect the print

nicedevil007:
Yeah that's what I mean, you can do it during a print, because KlipperScreen isn't affecting the print during a restart of its service.

PR #166, opened 2021-05-24
restart ks after icon theme change

alfrix:
#157


jordanruthe:
Can you make this a dialog prompting the user to confirm they want to restart immediately?

Issue #162, opened 2021-05-19
FR: Update all Stuff from moonraker that we can normaly update through the WebUI


nicedevil007:
Heya I guess everybody knows what I mean right?

Adding a button to just update the stuff from these parts would be awesome:

![image](https://user-images.githubusercontent.com/17103076/118847741-5abc5d80-b8ce-11eb-8661-f8f7d8a2da11.png)

jordanruthe:
Closing this since update manager capability has been merged

PR #183, opened 2021-06-10
Add update manager capability

jordanruthe:
This will address #162


panels/system.py
```diff
@@ -53,25 +57,211 @@ def initialize(self, panel_name):
...
+                self.labels[prog] = Gtk.Label("")
+                self.labels[prog].set_hexpand(True)
+                self.labels[prog].set_halign(Gtk.Align.START)
+
+                self.labels["%s_status" % prog] = self._gtk.Button()
+                self.labels["%s_status" % prog].set_hexpand(False)
+                self.labels["%s_status" % prog].connect("clicked", self.update_program, prog)
+                self.labels["%s_box" % prog] = Gtk.Box()
+                self.labels["%s_box" % prog].set_hexpand(False)
+                self.labels["%s_info" % prog] = self._gtk.ButtonImage("information",None, None, .7, .7)
```
alfrix, line 86: can this be "info" instead of "information" just to not add another very similar icon...

jordanruthe: Good catch. I don't know how I missed that icon

Issue #176, opened 2021-06-04
Feature Request: Show Last Console Log message


TravisWilder:
Hi
It would be create to see the last log lines of the klipper console on klipperscreen.
It would help like during QGL to see what is going on

jordanruthe:
This capability has been merged

Issue #121, opened 2021-03-31
Terminal panel


bsjoelund:
Is it possible to get a shortcut / icon to open Terminal and when closing go back?

jordanruthe:
This capability has been merged

PR #185, opened 2021-06-12
Add console panel

jordanruthe:
Some styling work still needs to be completed, but releasing this for testing.

Addresses #176 and #121

@alfrix If you have an icon for solarized, let me know. 

I also need to figure out a way to load custom colors for styles. I'm debating a JSON file (which could also include other data) in the theme directory.

I'm also debating if the colors should be customizable from the settings panel. Thoughts are welcome here. 

image

### DIFF
--- a/ks_includes/KlipperScreen.conf
+++ b/ks_includes/KlipperScreen.conf
@@ -118,6 +118,11 @@ icon: motor-off
 method: printer.gcode.script
 params: {"script":"M18"}
 
+[menu __main actions console]
+name: {{ gettext('Console') }}
+icon: console
+panel: console
+
 [menu __main config bedlevel]
 name: {{ gettext('Bed Level') }}
 icon: bed-level
--- a/ks_includes/KlipperScreen.conf
+++ b/ks_includes/KlipperScreen.conf
@@ -196,6 +196,11 @@ icon: custom-script
 panel: gcode_macros
 enable: {{ printer.gcode_macros.count > 0 }}
 
+[menu __print console]
+name: {{ gettext('Console') }}
+icon: console
+panel: console
+
 [menu __print network]
 name: {{ gettext('Network') }}
 icon: network
--- a/ks_includes/KlippyGtk.py
+++ b/ks_includes/KlippyGtk.py
@@ -11,6 +11,7 @@ klipperscreendir = os.getcwd()
 class KlippyGtk:
     labels = {}
     font_ratio = [51, 30]
+    keyboard_ratio = .22
     width_ratio = 16
     height_ratio = 9.375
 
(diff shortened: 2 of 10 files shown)


alfrix:
The svg icon for colorized
[console.zip](https://github.com/jordanruthe/KlipperScreen/files/6642962/console.zip)


problems at low res:
2021-06-12-191529_480x320_scrot
2021-06-12-192113_480x320_scrot

alfrix:
> I also need to figure out a way to load custom colors for styles. I'm debating a JSON file (which could also include other data) in the theme directory.
idk, perhaps 
> I'm also debating if the colors should be customizable from the settings panel. Thoughts are welcome here.

i thought in using two css files: a main shared one with everything 
and individual css files for themes that want to override values
that may be better from a maintainability standpoint.

jordanruthe:
What resolution are you running at? I'm unable to reproduce the keyboard looking like that at 840x480 and 480x320.

I will probably switch to a master css file with over-rides in the future. Unfortunately for the console, it isn't CSS, it's pango markup in a GTK TextView which supports the color field, but not classes

alfrix:
>What resolution are you running at?

480x320 (one of the cheap 3.5")

>I'm unable to reproduce the keyboard looking like that at 840x480 and 480x320.

i have just re-run the installer to get the new dependencies and even rebooted the pi, but still looks like that

alfrix:
it works with: 

```
--- a/screen.py
+++ b/screen.py
@@ -67,7 +67,7 @@ class KlipperScreen(Gtk.Window):
     files = None
     filename = ""
     keyboard = None
-    keyboard_height = 200
+    keyboard_height = 110
     last_update = {}
     load_panel = {}
     number_tools = 1
```
more and it starts repeating itself instead of getting bigger

alfrix:
2021-06-12-225559_480x320_scrot

jordanruthe:
> it works with:
> 
> ```
> --- a/screen.py
> +++ b/screen.py
> @@ -67,7 +67,7 @@ class KlipperScreen(Gtk.Window):
>      files = None
>      filename = ""
>      keyboard = None
> -    keyboard_height = 200
> +    keyboard_height = 110
>      last_update = {}
>      load_panel = {}
>      number_tools = 1
> ```
> 
> more and it starts repeating itself instead of getting bigger

The newest push makes the keyboard height dynamic and should resolve this.

alfrix:
now it's the correct size
the only thing i see that still needs fixing is that the cursor/selection markers are not transparent

jordanruthe:
> now it's the correct size
> the only thing i see that still needs fixing is that the cursor/selection markers are not transparent

I've been doing some research and trail/error, but I'm unable to find the right thing to make that block transparent. Let me know if you have any ideas.

alfrix:
apparently the block is transparent (changes according to the background), so there is a layer issue, the keyboard is not considered as background, furthermore that issue is present wherever the keyboard is used:
2021-06-16-122504_480x320_scrot

i have been using the console panel to quickly check the TUNNING_TOWER outputs, it's really useful
probably worth merging as is

PR #200, opened 2021-07-09
Add temp to job_status

alfrix:
2021-07-09-130728_480x320_scrot


alfrix:
I changed my mind on this, it's not much different from going to settings > temperature, and the status panel has a lot of stuff already

PR #202, opened 2021-07-09
Reorganize zcalibrate

alfrix:
Before
2021-07-09-151434_480x320_scrot
After
2021-07-09-151531_480x320_scrot

with this PR it looks more like the move panel


jordanruthe:
Thanks for your help with everything

PR #208, opened 2021-07-21
fixing xinput argument

peterus:
found a little "typo" about an xinput argument.


jordanruthe:
Thanks for catching this

Issue #216, opened 2021-07-28
Feature Request: Put User Selected Macros on Main Page


jallen810:
New (Former PanelDue) user here, great setup, really digging the UI but there is one feature I reeaaaaallly miss. Im running a bunch of macro's (Load/Unload Filament, QGL, etc.) and I would very much like to be able to just click them from the main screen without clicking thru a bunch of menus. I mocked up what I would like to see in the attached photo

![main_panel_MarkedUp](https://user-images.githubusercontent.com/62486440/127257576-4da244b0-c7d4-44cf-a524-8a7c203c789e.png)

jordanruthe:
We've integrated a macros shortcut.

PR #232, opened 2021-08-18
Macros shortcut

alfrix:
TODO: 
1. hide the shortcut on the splash screen
2. hide the config option if printers > 1

#216 

2021-08-18-181327_480x320_scrot


jordanruthe:
Can you make it so this is an option that can be turned on/off? Otherwise I like the idea.

jordanruthe:
Thanks!

Issue #235, opened 2021-08-21
Print screen alignment fails if filename to long


hawkeyexp:
Hi, if a filename is longer and an auto linebreak is done it could happen that the buttons on bottom moves out of screen and be only visible partially. In my case the filename split has used 5 lines where i noticed this behaviour.
I think a limit for linebreaks would be usefull or possible only showing the first line in big font and if a linebreak is needed the other lines in smaller fonz size.

alfrix:
A posible solution would be single line label (nowrap) and ellipsize end

alfrix:
Long filename tests, 2 lines seems to be the maximum
![2021-08-22-194436_480x320_scrot](https://user-images.githubusercontent.com/1247237/130372581-2197086b-9bbf-4bee-b765-87abf6fb0f3b.png)
![2021-08-22-194912_480x320_scrot](https://user-images.githubusercontent.com/1247237/130372582-3183c27e-3ff0-4124-b25b-40cf019a4f7c.png)
![2021-08-22-195104_480x320_scrot](https://user-images.githubusercontent.com/1247237/130372584-b1649de9-6b98-426b-a6ca-696908257277.png)

M117 also causes problems and should be ellipsized
![2021-08-22-194348_480x320_scrot](https://user-images.githubusercontent.com/1247237/130372579-d6363464-5df0-464b-9d13-781fbd529aa3.png)

hawkeyexp:
Looks good to me now

PR #238, opened 2021-08-23
Ellipsize Job_status Filename and LCD message

alfrix:
#235
2021-08-22-205903_480x320_scrot


jordanruthe:
What are you thoughts about possibly putting the file name in a scrolled window so it can be still viewed?

alfrix:
This PR is a quick fix for a problem.
I think that the ideal solution would be implementing scrolling labels like:
https://pyinmyeye.blogspot.com/2012/08/tkinter-animated-labels-demo.html

PR #245, opened 2021-08-28
Simple font size selector

### LINKED ISSUE #133: Changing the font size
Hey! Could you add the option to select the font size in the settings, since the standard font on a 4" screen seems very small to me, although the tiles and button icons are of a good size. Thank you!

alfrix:
Fix #133 
The difference is quite noticeable in all tested resolutions, here some samples from the two extremes

at 480x320 it ranges from 9px to 12px
2021-09-08-203659_480x320_scrot
2021-09-08-203736_480x320_scrot

and at 1080p ranges from 34px small  to 40px large
2021-09-08-203409_1920x1080_scrot
2021-09-08-203341_1920x1080_scrot

### DIFF
--- a/ks_includes/config.py
+++ b/ks_includes/config.py
@@ -134,6 +134,12 @@ class KlipperScreenConfig:
             {"24htime": {"section": "main", "name": _("24 Hour Time"), "type": "binary", "value": "True"}},
             {"side_macro_shortcut": {"section": "main", "name": _("Macro shortcut on sidebar"), "type": "binary",
                 "value": "True", "callback": screen.toggle_macro_shortcut}},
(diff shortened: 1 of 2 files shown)


alfrix:
Large +1 breaks the main menu if height < 365
2021-08-28-110324_480x320_scrot

alfrix:
1024x600

large
remmina_Pi_192 168 1 99_2021828-14531
small
remmina_Pi_192 168 1 99_2021828-145321

alfrix:
the top header bar may need to be adjusted according to aspect-ratio, to gain that little vertical space needed, on the small 3.5 it also is too small on 16:9 on 1080p

PR #246, opened 2021-08-28
Macro shortcut, ask to reload

alfrix:
I had to implement the callback for toggles


jordanruthe:
I'd rather KlipperScreen just remove or add the shortcut in on a callback. I'm happy to make this adjustment if you don't want to, but I might not get to it for a few days.

alfrix:
i'm not sure what you want to do, so i'll wait and see your commit :)

PR #248, opened 2021-08-30
Adjust titlebar height

alfrix:
1080p master
1080p_master
1080p This PR
1080p_PR
480x320  master
480x320_master
480x320 This PR
480x320_PR
1024x600 master
1024x600_master
1024x600 This PR
1024x600_PR


alfrix:
the white line comes from having the mouse on top of the extruder (hover highlight)

PR #270, opened 2021-09-13
Add string on POT and it

pappicio:


pappicio:
I don't know, conflict that cannt be solved in editor....

PR #273, opened 2021-09-17
Add temperature sensor readout to job status

kageurufu:


kageurufu:
Here is an example of how this displays

unknown-3

jordanruthe:
I'm going to be redoing the job panel shortly. Let's hold off on this until I can get that done as it may address what you want.

kageurufu:
Sounds good. I threw this together for myself, and figured I'd push it even just as a talking/kickoff point!

Thanks for the awesome software btw.

jbrandek:
it would be nice to add also layer that is in and from how many
somethin like Layer 12/200

alfrix:
the issue with this, is that it easily breaks the ui, another solution needs to be found
remmina_Pi_192 168 18 99_20220126-165123

kageurufu:
Yeah, after Jordan said the job panel would be redone I didn't bother working towards better layouts for that. 

I was also going to make the show sensors configurable, since I have way too many thermistors hooked up now and only want to see one.

PR #274, opened 2021-09-17
Allow gcode commands with preheat options

kageurufu:
I use this to run a heatsoak macro during preheat.
  It moves to bed center, maxes the part cooling fan,
  and starts my nevermore chamber filter to circulate
  more air.


jordanruthe:
Overall this looks good, can you also update the docs to include the gcode option?

kageurufu:
Done!

PR #277, opened 2021-09-18
some other fixes for ITA Lang

pappicio:
fixed and updated .po file, and updated .mo file for italian languge (if .mo is not the last, compiled with same .mo file, new fixes are hidden.)


alfrix (changes requested):
The 5 strings that have "#~" are not part of master, this means that you are using an outdated PO file

pappicio:
so, I have to use new pot file and recompile all parts?

alfrix:
always use the most recent version when doing a pr

pappicio:
ok, I'll provide to delete outdated strings, thanks.

PR #280, opened 2021-09-19
Z-bolt icon changes

alfrix:
Screenshot_20211128_143504


alfrix:
tell me if there is something that you don't like

jordanruthe:
I'd prefer to keep the extruder icons on the main page/extrude page the same as they currently are.

alfrix:
Updated screenshot with the icons side by side

PR #301, opened 2021-10-10
Add Input Shaper Panel

mikkogozalo:
Hello there,

I am tired of having to go back and forth my iPad or PC when calibrating using ADXL. That's why I created a new panel that can automate this process.

This is how it works. When panel is opened, it does ACCELEROMETER_QUERY to check if ADXL is setup correctly. If not, it will give you a message saying so. If it is setup, it will say whether to check for Faulty Wiring or start autocalibration. You can also manually change the runtime settings for input shaper within the panel.

If you autocalibrate, it will just call `SHAPER_CALIBRATE` and do a `SAVE_CONFIG` afterwards.

No ADXL attached or faulty wiring detected
Screenshot_20211010_181723

ADXL attached, before clicking Auto-Calibrate and Save
Screenshot_20211010_181840

While calibrating
Screenshot_20211010_181916

### DIFF
--- a/panels/input_shaper.py
+++ b/panels/input_shaper.py
@@ -1,6 +1,5 @@
 import gi
 import re
-import subprocess
 
 gi.require_version("Gtk", "3.0")
 from gi.repository import Gtk, Pango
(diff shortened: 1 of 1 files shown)


moelski:
I would love to test this since my next step is input shaper calibration. 
Is this topic still under development? And is there any information how to get this panel running?

PR #305, opened 2021-10-17
Update Configuration.md

Drachenkaetzchen:
Updating the documentation with more verbose instructions about the `KlipperScreen.conf` file and what to do if it doesn't get copied automatically.

Background: Earlier today, a user was following the guide and they were unable to find any `KlipperScreen.conf` file on their system, despite doing a search on the full filesystem and the documentation suggesting that the file will always be copied.


jordanruthe:
I do agree, the wording is confusing. I don't like suggesting that the default config should be copied though, as that will cause issues for most users after an update (they would need to manually re-copy the config to get a new version). Would you mind updating the wording to suggest creating a new, empty file that they can add the necessary configuration items to?

PR #309, opened 2021-10-19
Fix german translation

hawkeyexp:


jordanruthe:
Can you run the `update_translation_files.sh` and include that in this PR so the binary file is updated?

hawkeyexp:
Hm, seems with latest commits the german translation has not been fixed - klipperscreen is still showing "Last" instrad of "Laden" - possible a leftover hardcoded part in the gui  (Filament change) ?!

Elektrickser:
Hi Hawkeyexp,

I searched the msgid, but don't fount it in the code. In which panel you have this?

hawkeyexp:
It's in the extrude panel if load and unload macro are available and the additional buttons are shown

Elektrickser:
Wait a moment, i will make the current changes.
I see it.

PR #310, opened 2021-10-19
Configurable xy position for z-calibrate added

ckvsoft:
Configurable xy position for z-calibrate added

Example for the KlipperScreen.conf
[z_calibrate_position]
calibrate_x_position: 100
calibrate_y_position: 100

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -62,6 +62,15 @@ moonraker_api_key: False
 z_babystep_values: 0.01, 0.05
 ```
 
+## Z probe calibrate Option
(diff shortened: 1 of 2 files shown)


jordanruthe:
I'm not against this, but there does need to be a check for safe_z_home. If safe_z_home exists, it should prefer that rather than a default config of 0, 0. I'm not sure if there's a way to specify the location in the klipper config, but that might be preferred if it's possible.

jordanruthe:
Fair point. Can you update the Configuration.md with your new options? Besides that I think this is good to merge

ckvsoft:
Yes, I'll do it.
lg Chris

PR #327, opened 2021-11-12
Zcal panel: Support for Z_ENDSTOP_CALIBRATE

### LINKED ISSUE #102: Z-Probe Calibration
If we do Z-Probe Calibration with KlipperScreen the calibration is started without checking if the printer is really homes. F.e. If you homed your printer in the past and then moved the printhead out of this position for anything else (filament change f.e.) the z-calibrate will start at this not centered state! That can damage the printbed while the probe is outside the bed during this movement.

Maybe it is possible to add a check if the printer is at its homed position or just let it home every time one is clicking on the probe calibration procedure.

### LINKED ISSUE #236: Z Calibrate panel and Voron printers
Z Calibrate panel call PROBE_CALIBRATION gcode but on Voron printers should call/manage Z_ENDSTOP_CALIBRATE.
Is it possible to add an option to choose between PROBE_CALIBRATE and Z_ENDSTOP_CALIBRATE ?

alfrix:
Also now it doesn't start until the button is pressed
i have left the abort button always active in case the calibration was started in the web interface, in that case you can just press start and continue with your calibration or abort.

Should Fix  #236 and also Fix #102
2021-11-11-235737_480x320_scrot
2021-11-11-235853_480x320_scrot


panels/zcalibrate.py
```diff
@@ -90,7 +92,19 @@ def activate(self):
...
             if x_position > 0 and y_position > 0:
                 self._screen._ws.klippy.gcode_script('G0 X%d Y%d F3000' % (x_position, y_position))
-        self._screen._ws.klippy.gcode_script(KlippyGcodes.PROBE_CALIBRATE)
+
+        if (self._printer.config_section_exists("probe") or self._printer.config_section_exists("bltouch")):
+            self._screen._ws.klippy.gcode_script(KlippyGcodes.PROBE_CALIBRATE)
+        else:
+            self._screen._ws.klippy.gcode_script(KlippyGcodes.Z_ENDSTOP_CALIBRATE)
+
+        self.start.get_style_context().remove_class('color3')
```
jordanruthe, line 101: Can you change these to be in the self.labels dictionary? It may be inappropriately named at this point, but it's where widgets are stored. I can update the name in the future to self.widgets

panels/zcalibrate.py
```diff
@@ -58,30 +67,23 @@ def initialize(self, panel_name):
...
+        grid.attach(pos, 1, 1, 1, 1)
+        grid.attach(self.zneg, 0, 1, 1, 1)
+        grid.attach(self.complete, 2, 0, 1, 1)
+        grid.attach(distances, 0, 2, 3, 1)
+        grid.attach(cancel, 2, 1, 1, 1)
 
         self.content.add(grid)
 
-    def activate(self):
+    def start_calibration(self, widget):
```
jordanruthe, line 86: I believe it'd be necessary to track whether or not a start was triggered. The case I'm thinking about is if a user hits start, then backs out of the panel without aborting the command.

alfrix: The status of the panel remains, you can back-out and re-enter the panel and it will be left as it was

jordanruthe: I'm good with the abort button being always available, but I'm more worried about a case where the printer gets stuck in a weird state due to the user backing out of a panel and then proceeding to do other things without the calibration being properly aborted. It is a change from the current panel behavior.

alfrix: It is not a change from current panel behavior, the current panel does not abort when going back, it behaves exactly the same but with the non optional start, i haven't noticed it being an issue, but i can add new functionality to abort on panel back

jordanruthe: Actually, you are correct. For some reason I thought it had been.

PR #338, opened 2021-11-24
Improve the bed mesh plot

alfrix:
Before:
colorized_before
mdark_before
zbolt_before

After:
colorized_after
mdark_after
zbolt_after


jordanruthe:
How does this handle the case of a point being greater than .5?

alfrix:
Artificially made the value +5 and -1,
Master:
before
This PR
after

alfrix:
With the last patch:
big_small
offset
verywarped

PR #340, opened 2021-11-25
Wakeup touch block

### LINKED ISSUE #207: [possible solution] enable buttons after screen wakes up
hi all, 

maybe somebody already figured this out but when i klick on the touchscreen to wake it up it also selects buttons on the screen. 
it now happend that i woke the screen up by touching it directly at the place where the emergency stop was wich in term ment that klipper got restarted.

could it be possible to wake up on touch and after x milliseconds enable the buttons so that it would be less lickely u would wake up the touchscreen and klick on buttons with one touch?

### LINKED ISSUE #292: Pushing buttons while screen is timed out
Hi,
I can push buttons when screen is timed out. How do I disable this functionality?
I just wanted to wake the screen and did a G28 :)

alfrix:
Should fix #207, fix #292, resolve #242


screen.py
```diff
@@ -546,12 +547,23 @@ def remove_subscription(self, panel_name):
...
+        if state == functions.DPMS_State.Off and visible:
+            logging.info("DPMS State Off -> Hiding")
+            self.hide()
+            os.system("xsetroot  -cursor_name  watch")
+            self.touch_ready = False
+        elif state == functions.DPMS_State.On and not visible:
+            if self.touch_ready:
+                logging.info("DPMS State On -> Showing KlipperScreen")
+                self.show()
+                os.system("xsetroot  -cursor_name  arrow")
```
jordanruthe, line 561: This shouldn't necessarily be arrow. If you look at line 143, it could be arrow or blank.

alfrix: xsetroot only affects the root window, the cursor definitions that you mention are respected inside KlipperScreen but not during hide (that's why you can see the watch cursor), the arrow or watch that are in here will be shown on windows that are not KS, and no unfortunately with this method you cannot set it blank, that needs to be done during xinit using the -nocursor flag, xsetroot can only set names in /usr/include/X11/cursorfont.h

alfrix: it that there is a very old workaround of using an xbm, i will try that

juulius: hi guys, just saw the changes made in the code. im kinda hornored, enthusiastic you guys took the proposal and run with it :D 

just wanna pass some extra info. 
i made some extra logging in my version for logging after x time and noticed that when the screen is blank the state is suspend and not off (one off the main reasons that i used not on)
log:
`2021-11-30 00:09:03,604 [screen.py:check_state()] - log line after checking 600.421985 seconds:  checked 1199 times, DPM_state Suspend, window is hidden`


(just for info the code that makes the log)
```
        state = functions.get_DPMS_state()
        global counter, time_start, count
        log_text = ["FAIL", "ON", "Standby", "Suspend", "OFF"]
        log_visible =["hidden", "visible"]
        isvisible = self.get_property("visible")
        time_interval = time.monotonic()
        if time_start == 0.0:
            logging.info("First run, checked %d times, DPM_state %s, window is %s", count, log_text[state + 1],
                         log_visible[isvisible])
            time_start = time_interval
            count = 0
        elif (time_interval - time_start) > 600.0:
            logging.info("log line after checking %f seconds:  checked %d times, DPM_state %s, window is %s",
                         time_interval - time_start, count, log_text[state + 1], log_visible[isvisible])
            time_start = time_interval
            count = 0
```

juulius: no problem, i didn't play with the timeout value yet. i've got it set to 500 now, will test 200 and lower

juulius: Hi, i tested some values, lower then 200 introduces the touch issue again (for me). this also may be personal, some maybe want a more responsive system others not

just a quote from #207 
> have already seen the "status information". it was helpful for testing, but that's true. this will unnecessarily flood the journal. I commented it out
> 
> I also played around with the delay-time. 500ms seems good to me (250ms is to quick for me, misclicks happens)

i think for now 500 ms is fine enough

juulius (commented):
maybe im a little whiny, but the state variable only will give a number  in de log text
`logging.info("DPMS State %s -> Hiding", state)` 

i personally find it neater to have text 
i changed mine to this
```
 def check_dpms_state(self):
        state = functions.get_DPMS_state()
        Log_DPMS  = ["FAIL", "ON", "Standby", "Suspend", "OFF"]
```
`       logging.info("DPMS State %s -> Hiding", log_DPMS[state+1])`

for the rest the fix is working great, the loading symbol for the pointer is a neat detail.

PR #359, opened 2021-12-08
add speed as parameter to UN/LOAD_FILAMENT macro

ckvsoft:
Hallo

So we can now transfer the speed from the extrude panel to the macro.
The macros were ok, only when I wanted to load the TPU it was too fast.

example macros:
[gcode_macro LOAD_FILAMENT]
gcode:
    {% set speed = params.SPEED|default(500) %}
    G91
    {action_respond_info("SPEED= : %s" % (speed))}
    G1 E50 F{speed}
    G1 E50 F{speed}
    G92

[gcode_macro UNLOAD_FILAMENT]
gcode:
    {% set speed = params.SPEED|default(500) %}
    G91
    {action_respond_info("SPEED= : %s" % (speed))}
    G1 E-50 F{speed}
    G1 E-50 F{speed}
    G92


alfrix:
this could be interesting to tweak the purge speed, this would be my Macro:

[gcode_macro LOAD_FILAMENT]
gcode:
    {% set speed = params.SPEED|default(300) %}
    {% set max_velocity = printer.configfile.settings['extruder'].max_extrude_only_velocity %}
    M300 # beep
    G91 
    G92 E0
    G1 E350 F{max_velocity}
    G1 E25 F{speed} #purge
    M300
    M300

jordanruthe:
Thanks for this addition. Would you be able to add somethings to the docs detailing that speed can be passed to the LOAD/UNLOAD macro and including an example macro?

ckvsoft:
That is of course possible. I just didn't know what to do with it.
Should I create my own file or should I add it to panel.md (Extrude)?
Until recently it wasn't clear to me that I needed a load / unload macro to display this in the panel. Is that somewhere? Unfortunately I can't find anything.

jordanruthe:
I think it might need to be an added page to the documentation as there is currently no reference to those macros and it doesn't really fit into Configuration, as that's more pertaining to KlipperScreen's config. Maybe a new page such as Klipper Configuration or Supported Macros

ckvsoft:
Is the new macros.md file OK? Hope that fits

jordanruthe:
Thanks for the documentation update.

PR #368, opened 2021-12-12
Improve pathing

### LINKED ISSUE #366: KlipperScreen assumes that current working directory is source root
There are number of locations in the code that do something in the lines of:

```python
klipperscreendir = os.getcwd()
```

This and based on that value strings are constructed with lines similar to these:

```python
"%s/styles/%s/images/%s.svg" % (klipperscreendir, self.theme, str(image_name))
```

There are a few issues with this, since one could start KlipperScreen from outside of the source directory, the paths generated by these lines will point to nonexistent files. Which will cause KlipperScreen to fail.

A better solution would be to replace path detection with an implementation from `pathlib` and `os.path.join`, similar to this:

```python
klipperscreendir = pathlib.Path(__file__).parent.resolve()
os.path.join(klipperscreendir, "styles", self.theme_name, "images", str(image_name) + ".svg")
```

This will fix 2 issues. One is the reliance on current working directory as location for images, styles, locales and configuration defaults. The second is possible portability issues caused by different directory delimiter characters.

alfrix:
closes #366

### DIFF
--- a/ks_includes/KlippyGtk.py
+++ b/ks_includes/KlippyGtk.py
@@ -2,11 +2,12 @@
 import gi
 import logging
 import os
+import pathlib
 
 gi.require_version("Gtk", "3.0")
 from gi.repository import Gtk, Gdk, GdkPixbuf, Gio, GLib, Pango
 
-klipperscreendir = os.getcwd()
+klipperscreendir = pathlib.Path(__file__).parent.resolve().parent
 
 class KlippyGtk:
     labels = {}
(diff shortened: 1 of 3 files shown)


ressu:
Thanks for getting started on this, I think there were some occurrences of cwd in `ks_includes/config.py` too

https://github.com/jordanruthe/KlipperScreen/blob/9660172e0b5a1f029b39b0371959c589c69b4269/ks_includes/config.py#L32

PR #378, opened 2021-12-20
Rework Time estimates

### LINKED ISSUE #108: Remaining print time is always completely of
While printing the remaining time shown by Klipper screen does not match what Maintail shows. I tried with "file time" setting and "filament" setting, neither is near the Maintail values (which pretty much match the print).

Compare these photos, take within a minute. Especially, the percentage value is totally wrong.

Filament:
IMG_2429

File:
IMG_2430

Maintail:
IMG_2431

### LINKED ISSUE #222: Print time estimate is behaving very strange...
Whenever I start a print, the time-left estimate is way off. It starts ok-ish with time-left estimate close to what's expected and what Fluidd is showing. 

But then it starts to climb very rapidly. It's not uncommon for a small print, taking around 30 minutes, to show 200+hours on the estimate after a minute of printing.

After a while this way too high estimate starts dropping and returns back to a normal value.

### LINKED ISSUE #282: Job_status getting out of screen
Like #190 happens when the resolution is low and the font is large

Here it barely fits
2021-09-20-142938_480x320_scrot

And no longer fits
2021-09-20-143012_480x320_scrot

Proposal 1:
Remove the labels "Elapsed" and "Left"

Proposal 2
Remove the time estimation from elapsed

alfrix:
It should be much more accurate
Fixes #222
Fixes #108 
Fixes #282

### DIFF
--- a/ks_includes/KlippyGtk.py
+++ b/ks_includes/KlippyGtk.py
@@ -333,9 +333,14 @@ class KlippyGtk:
     def formatTimeString(self, seconds):
         time = int(seconds)
         text = ""
+        if int(time/86400) != 0:
+            text += str(int(time/86400))+"d "
         if int(time/3600) != 0:
-            text += str(int(time/3600))+"h "
-        text += str(int(time/60) % 60)+"m "+str(time % 60)+"s"
+            text += str(int(time/3600) % 24)+"h "
+        if int(time/60) != 0:
+            text += str(int(time/60) % 60)+"m "
+        else:
+            text = str(time % 60)+"s"
         return text
 
     def formatTemperatureString(self, temp, target):
--- a/ks_includes/config.py
+++ b/ks_includes/config.py
@@ -124,8 +124,8 @@ class KlipperScreenConfig:
             {"print_estimate_method": {
                 "section": "main", "name": _("Estimated Time Method"), "type": "dropdown",
                 "value": "file", "options": [
-                    {"name": _("File") + " " + _("(default)"), "value": "file"},
-                    {"name": _("Duration Only"), "value": "duration"},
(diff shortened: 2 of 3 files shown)


hapklaar:
Curious to know: does it use the same logic as Fluidd for example? That is usually a lot more accurate than slicer (Cura) for me.

alfrix:
i haven't looked at fluidd yet, this is not finished, i just made the pr for Jordan to see it and stop me if he doesn't want it

jordanruthe:
Thanks for looking into revamping this. What's the reason for getting rid of file and filament estimation methods?

alfrix:
> Thanks for looking into revamping this. What's the reason for getting rid of file and filament estimation methods?

File_time is extremely inaccurate at the start, it's only reasonable at about 10% to 15% and i suspect it's the reason for all the complaints, being the default everyone sees this issue. and it tends to always overestimate.

Filament_time it's a bit faster to reach convergence, but also could be a source of complaining that it's not accurate at the beginning and it tends to underestimate.

Slicer_time it's usually the most accurate during the first % and even overall if the user adjusts the slicer to produce the most accurate timing possible, i added the new parameter of compensation to help with that.

Based on those observations, it's best to move people to more accurate methods, so i created the auto setting, that seem to produce the best overall results, but it tends to jump around a bit (seconds), again to lessen potential sources of complaints i hided the seconds until the very end.

> does it use the same logic as Fluidd for example?

i have looked at Fluidd and Mainsail, Fluidd uses the speed_factor, which seemed a great idea, until i tested and off course my printer didn't react that linearly, so i had to adjust it. the rest is very similar.
Mainsail creates the estimation by averaging both Filament and File, and again it's terrible for a while until it converges

adawalli:
Any word on getting this merged?

PR #384, opened 2021-12-27
Makes "Confirm Emergency Stop" setting localizable

orobardet:
The string "Confirm Emergency Stop" in the configuration options was not translatable.  
This PR adds it into the PO file and the en translation files.

It does not update the other languages. I'm about to push a PR for fr_FR, with this new translation en some other changes to french language.


jordanruthe:
I think this was merged with #385. Feel free to re-open if needed

PR #402, opened 2022-01-05
Zcalibrate: Track the status of calibration and inform the user more

### LINKED ISSUE #409: zcalibrate.py accept(): misslng final line
After clicking **Accept** from the z calibrate panel, KlipperScreen remains there.  The `ACCEPT` command is sent to Klipper and can be seen in the console.

It looks like the last line in the accept() function is missing.  Appending the last line from the abort() function

`self.menu_return(widget)`

fixes the issue.

### LINKED ISSUE #410: Z Calibrate Accept Button Does Nothing (Visually)
When using the Z Calibrate function in Klipperscreen, the Accept button does not dismiss the calibrate screen or show any messages.  I have confirmed that it saves the value, but there is no feedback on this.

alfrix:
The buttons now activate and deactivate properly depending on the machine status.

The start button now has an identifier that informs what command it will send: 
probe for probe_calibrate, 
endstop for z_endstop_calibrate 
mesh bed_mesh_calibrate

There are now popups with errors or info to maintain the user informed
Closes #409 (can't do this breaks manual mesh calibrations and hides errors)
 Fixes #410

### DIFF
--- a/panels/zcalibrate.py
+++ b/panels/zcalibrate.py
@@ -36,9 +36,15 @@ class ZCalibratePanel(ScreenPanel):
         self.widgets['zpos'].connect("clicked", self.move, "+")
         self.widgets['zneg'] = self._gtk.ButtonImage('z-closer', _("Lower Nozzle"), 'color1')
         self.widgets['zneg'].connect("clicked", self.move, "-")
-        self.widgets['start'] = self._gtk.ButtonImage('resume', _("Start"), 'color3')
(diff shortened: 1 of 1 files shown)


moelski:
Why is it not possible to select which calibration should be done. My Voron has a probe and a Z-Endstop. I need to calibrate the endstop but KS always selects Probe ... Even MainSail asks you where to save your new calibration.

PR #411, opened 2022-01-08
Increase max speed and separate XY and Z max speed

randellhodges:
This change will increase the maximum speed for XY moves to 500 and gives an option to set the maximum speed for Z moves.

Ideally the maximum speed would be read from Klipper `max_velocity` and `max_z_velocity`. Maybe as an enhancement to this change.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -51,6 +51,12 @@ moonraker_api_key: False
 
 # Define the z_babystep intervals in a CSV list. Currently only 2 are supported
 z_babystep_values: 0.01, 0.05
+
+# Override the movement speed and set a specific for this printer.
+# These setting overrides the settings configured in the UI. If specified,
+# the values configured in the UI will not be used.
+move_speed_xy: 500
+move_speed_z: 300
 ```
 
 ## Z probe calibrate Option
--- a/ks_includes/config.py
+++ b/ks_includes/config.py
@@ -114,9 +114,12 @@ class KlipperScreenConfig:
                           "callback": screen.restart_warning, "options": [
                               {"name": _("System") + " " + _("(default)"), "value": "system_lang"}
             ]}},
-            {"move_speed": {
-                "section": "main", "name": _("Move Speed (mm/s)"), "type": "scale", "value": "20",
-                "range": [5, 100], "step": 1}},
+            {"move_speed_xy": {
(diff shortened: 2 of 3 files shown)


jordanruthe:
I think the max speed of 500mm/s seems a bit excessive. I'm fine with raising it above 100, but realistically most printers wouldn't be able to even reach that and it will make the lower limits harder to select.

randellhodges:
If you don't want the UI to have a max, which I can understand with a
slider, can we be able to manually set a speed in the config file?

With 48v steppers, I can easily get 500mm/s travel moves (with an 8k
acceleration). I'm not sure it makes it to 500 but it sure makes a
difference in the moves.

jordanruthe:
I have no issues with removing the maximum and allowing it to be 500 if you manually set it in the config. I think the UI max should likely be 200.

randellhodges:
I changed the upper limit to the speed to 200.

In addition, I made a minor change so you can tweak the value on each printer definition. I also updated the documentation and included a note on how that value would override the UI.

alfrix:
Is there a reason for not using printer max_velocity as the maximum value?

randellhodges:
> Is there a reason for not using printer max_velocity as the maximum value?

Nope, I didn't think about it when I made this last set of changes. When using the move panel, we have a connection to the printer, so we could use that value.

The settings panel, I didn't know how to implement those changes. In my opinion, the best way would be to use the max velocity on a per printer basis when you are in the config/settings panel and then store it on the printer section. The only thing is that I'm not sure how to do a per printer settings on that panel. I can look if that is what you are referring to.

randellhodges:
Would you think moving it to the limits panel be the best place to put it since it would be a printer-by-printer setting? Hmm, but it would be confusing since it isn't really a limit like the other values there.

alfrix:
imho, the ideal place for the setting is the move panel,  i'm thinking in a button that changes the dist buttons to the slider

randellhodges:
I played around a bit with the idea of flagging configuration options as printer specific. It started to sprawl into some much larger changes in how confg and printer specific values are read and set. I decided to roll back those changes.

If we want to support printer specific settings set by the UI, I'd propose that would be split out into its own change. For now we just go down the lines of reading the value from the printer section if specified?

PR #417, opened 2022-01-08
Fix zcalibrate.py accept()

### LINKED ISSUE #409: zcalibrate.py accept(): misslng final line

randellhodges:
fix #409

### DIFF
--- a/panels/zcalibrate.py
+++ b/panels/zcalibrate.py
@@ -133,6 +133,7 @@ class ZCalibratePanel(ScreenPanel):
     def accept(self, widget):
         logging.info("Accepting Z position")
         self._screen._ws.klippy.gcode_script(KlippyGcodes.ACCEPT)
+        self.menu_return(widget)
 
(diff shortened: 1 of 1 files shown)


alfrix:
This breaks manual bed mesh calibration
and because there can be errors on accept it will break #402 information system

randellhodges:
Good catch. I remember now a similar conversation when I did a PR for
something in Klipper.

randellhodges:
Closing this because as pointed out by @alfrix, it only addresses a "happy path" for automatic bed mesh calibration, it doesn't cover manual bed mesh or errors during the bed mesh processes.

PR #418, opened 2022-01-10
Use Custom Service

Devil-of-Chaos:
to restart xrdp and don't user KlipperScreen.service


jordanruthe:
Can you update the documentation to include the new option under main?

Devil-of-Chaos:
I hope the explanation is ok

PR #445, opened 2022-01-15
Bed leveling fix for printers with 4+ bed_screws

### LINKED ISSUE #443: Bed level broken with 6 screws in the config
I was setting up my Tronxy X5SA-400-Pro that has 6 screws in its config. The bed level was behaving really weird, going in a weird triangle, and not a square.

apolkosnik:
Fixes #443 on my printer with  6 bed screws.
It seemed to work with 3, 4, and 6 bed screws defined in printer.cfg
Also, added if statement to check for "screws_tilt_adjust" to avoid throwing an error when the section is not present.

### DIFF
--- a/panels/bed_level.py
+++ b/panels/bed_level.py
@@ -74,19 +74,27 @@ class BedLevelPanel(ScreenPanel):
                 [x, y*3],
                 [x*3, y*3],
             ]
-            logging.debug("Calculated screw locations [x,y]: %s", screws)
+            logging.debug("Calculated screw locations [x,y]: %s", self.screws)
         else:
-            logging.debug("Configured screw locations [x,y]: %s", screws)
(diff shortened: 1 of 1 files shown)


alfrix:
This is wrong
```
        if self._printer.config_section_exists("screws_tilt_adjust"):
            self.labels['screws'].set_sensitive(True)
```
the button will not exist:
```
        if self._printer.config_section_exists("screws_tilt_adjust"):
            self.labels['screws'] = self._gtk.ButtonImage("refresh", _("Screws Adjust"), "color4")
            self.labels['screws'].connect("clicked", self.screws_tilt_calculate)
            grid.attach(self.labels['screws'], 3, 0, 1, 1)
```

alfrix:
Tested, it's seems ok, there is potential to add the extra screws if you do the calculation for missing 4 positions i will do the ui changes to integrate them

PR #454, opened 2022-01-18
Minor corrections

srbjessen:
Slight changes in translations and wordings

I'm missing the translation ID and String for "hour" and "hours" in "Screen timeout" under settings. 
They don't appear to be available in the 'en'-locale either?


alfrix:
This PR is conflicting with master, you should close it, grab the po file from the last version in master and do your updates with the new file and then open a new PR

PR #465, opened 2022-01-22
da-translation

srbjessen:
Not sure what the problem was with the previous PR, but this should work


alfrix:
Now it looks good, but missing the .mo you can generate it from poedit

PR #468, opened 2022-01-23
Polish translation

apolkosnik:
PR contains Polish translation


alfrix:
Some labels are too long, try to respect the original string lengths, be reasonable
a

apolkosnik:
I'll take a look, thanks for the feedback

apolkosnik:
Ok, I'm testing the updated translation. I've noticed that "hour" and "hours" are not in the .pot file.

apolkosnik:
I've played with Poedit a little, and it looks like "hour" can have more than two forms in some languages, I've also updated the KlipperScreen.pot file. Other language translations might need to be updated to have "hour" translated in the dropdown for the "Screen Power Off Time"

alfrix:
The pot file its autogenerated, the script will erase your changes when it runs, plurals need to be added in the py files with `ngettext` or something like that, and the script needs to be updated to accept plurals `keyword=_:1,2`...

The translation looks ok in the ui now, there is a minor issue with the system panel that i will fix

PR #507, opened 2022-02-13
base_panel: added light toggle button where back button is used to be…

Flo-Kra:
… whenever no back button is displayed (only if a power device starting "light" in it´s name is configured. 
Does not affect original back-button behaviour. 

KlipperScreen_LightToggleButton
)


AntoszHUN:
Now is it possible to adjust the brightness then?

Flo-Kra:
> Now is it possible to adjust the brightness then?

No, as this button is only a shortcut so that the light can be toggled with one click instead of 3 or 4, regardless if currently printing or not. It uses the same backend functions as the light switch already available in the power menu. 
As it relies on functions of Moonraker (power device with name "light" configured in moonraker.conf) it can only do what´s supported there.

AntoszHUN:
> 

I understand. Thanks for the reply!

alfrix:
i simply do not agree with replacing navigation buttons with other actions.

Issue #518, opened 2022-02-21
[FR] Switch power device on with touchscreen wakeup


hawkeyexp:
It would be nice to define a moonraker power device which will be switched on if touchscreen was in sleep (and this power device is off) and you tap on it to wake it up. This would allow a full wakeup of the system if klipper is powering of a moonraker device after a timeout while not printing (for example in my case the 24V line will switch off after a print with timeout and the pi keeps running).

PR #533, opened 2022-03-06
screen: power on screen wakeup

alfrix:
#518


CornHead764:
Could an extension of this pull request also include the ability to shutdown devices on sleep? essentially the opposite of power on wake.

alfrix:
yes, you can i will not put any more efforts on this or any power_device related thing, my testing using a klipper output_pin as a power_device doesn't seem to translate to other type of devices as seen with the issues experienced by other users

alfrix:
yes, you can i will not put any more efforts on this or any power_device related thing, my testing using a klipper output_pin as a power_device doesn't seem to translate to other type of devices as seen with the issues experienced by other users

PR #560, opened 2022-03-25
Update Troubleshooting.md

pthompson8594:
Seems the .conf has been changed to .config


alfrix:
i was faster b7d58a7ffb555979832f25fc035e17c7b996b68c
but thank you for finding out

PR #579, opened 2022-04-06
Add possibility to define custom code for cooling

### LINKED ISSUE #578: [Enhancement] Add custom gcode for "cooldown" preset
Hi!

I like the ability to run custom gcode after using a temperature preset. I would love it even more if this could be done for the _cooldown_ preset as well.

Something along the line:

```
[preheat cooldown]
gcode: 
    TURN_OFF_HEATERS
    M107

[preheat ABS]
bed: 110
extruder: 245
gcode: 
    HEATSOAK T=110 MOVE=1
```

This is something I found possible in mainsail:
Screenshot 2022-04-07 at 01 47 32

Note that the _cooldown_ preset could be just another preset and treated as such in the code.

I anyway have implemented a quick change in a fork of mine.

BTW Thank you for this great work that is **KlipperScreen**

kohlerj:
Linked to issue #578 

These changes allow to define custom code to be run when using the _cooldown_ preset. It makes the _cooldown_ preset similar to the _other_ ones. Note that this is something that can be done on Mainsail.

Here an example in the configuration:
```
[preheat cooldown]
gcode: 
    TURN_OFF_HEATERS
    M107
```

Note that the cooldown could be made a preset like any other? I'd be happy to take a look at it if wanted!

fixes #578

### DIFF
--- a/panels/temperature.py
+++ b/panels/temperature.py
@@ -83,9 +83,10 @@ class TemperaturePanel(ScreenPanel):
     def preheat(self):
         self.labels["preheat_grid"] = self._gtk.HomogeneousGrid()
         for i, option in enumerate(self.preheat_options):
-            self.labels[option] = self._gtk.Button(option, "color%d" % ((i % 4)+1))
-            self.labels[option].connect("clicked", self.set_temperature, option)
-            self.labels['preheat_grid'].attach(self.labels[option], (i % 2), int(i/2), 1, 1)
+            if option != "cooldown":
+                self.labels[option] = self._gtk.Button(option, "color%d" % ((i % 4)+1))
+                self.labels[option].connect("clicked", self.set_temperature, option)
+                self.labels['preheat_grid'].attach(self.labels[option], (i % 2), int(i/2), 1, 1)
         scroll = Gtk.ScrolledWindow()
         scroll.set_property("overlay-scrolling", False)
         scroll.set_hexpand(True)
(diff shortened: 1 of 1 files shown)


kohlerj:
Please close this if inappropriate.

alfrix:
i do see a benefit here, turning off the part fan, changing leds colors, etc

you might already seen this, but i will point out that you would lose functionality when using TURN_OFF_HEATERS, the reason is that in KS you can select the heaters to be cooldown, and that isn't possible if you run that Gcode

i will also point out that you can share the same Gcode between interfaces, just create a macro and enter the macro name in both, it would be easier to maintain if you decide to change the Gcode later

kohlerj:
> you might already seen this, but i will point out that you would lose functionality when using TURN_OFF_HEATERS, the reason is that in KS you can select the heaters to be cooldown, and that isn't possible if you run that Gcode

Well it would be up to the user if she/he wants to loose that functionality. Custom gcode is her/his responsability...

> i will also point out that you can share the same Gcode between interfaces, just create a macro and enter the macro name in both, it would be easier to maintain if you decide to change the Gcode later

Didn't quite catch that one, could you elaborate? The gcode I pasted is an example. Obviously in the default.conf with would find an empty "cooldown" gcode...

alfrix:
> Didn't quite catch that one, could you elaborate?
point all the Uis to the same macro, for example "_preheat_PLA" so you can change the temp in the macro and all the UIs will be updated, just a tip

> Obviously in the default.conf with would find an empty "cooldown" gcode...
then why you didn't add that to the PR

alfrix:
i added it to the docs instead of default.conf 2c6a853a80285490d7d61bab1f8a9a44d2b3f620

kohlerj:
Thanks for considering my request and thank you for modifying the docs. 💯

PR #670, opened 2022-07-09
Tr_update#6

technical50:
Update's


alfrix:
when a word is abbreviated, please keep it that way

PR #708, opened 2022-08-21
Add current heater power % to job status

lalten:
I thought it would be nice to see where I'm at with heater cartridge heating capacity, so I added the current heater power % to job status.

image

### DIFF
--- a/ks_includes/config.py
+++ b/ks_includes/config.py
@@ -229,7 +229,8 @@ class KlipperScreenConfig:
                 "range": [50, 150], "step": 1}},
             {"autoclose_popups": {"section": "main", "name": _("Auto-close notifications"), "type": "binary",
                                   "value": "True"}},
-
+            {"show_heater_power": {"section": "main", "name": _("Show Heater Power"), "type": "binary",
+                                   "value": "False"}},
             # {"": {"section": "main", "name": _(""), "type": ""}}
         ]
 
--- a/ks_includes/config.py
+++ b/ks_includes/config.py
@@ -117,7 +117,8 @@ class KlipperScreenConfig:
             if section == 'main':
                 bools = (
                     'invert_x', 'invert_y', 'invert_z', '24htime', 'only_heaters', 'show_cursor', 'confirm_estop',
-                    'autoclose_popups', 'use_dpms', 'use_default_menu', 'side_macro_shortcut', 'use-matchbox-keyboard'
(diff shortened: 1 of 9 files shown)


lalten:
There was an issue with other screens (`update_temp() missing 1 required positional argument: 'power'`) and the linter was complaining about one line being too long. force-pushed an update.

alfrix:
nice, i sometimes hit the limit when printing fast on a volcano, this would have been useful.

it needs to be optional, and as it takes space from the possible optional extra heaters/sensors the code that adds the extra items would need to be modified, and tested on lower resolutions.

lalten:
Added a `show_heater_power` config option and moved all the update_temp() functions to the parent class.

panels/job_status.py
```diff
@@ -186,7 +186,7 @@ def create_status_grid(self, widget=None):
                 self.heater_button[device].set_halign(Gtk.Align.START)
                 self.labels['temp_grid'].attach(self.heater_button[device], n, 0, 1, 1)
                 n += 1
-        extra_item = True
+        extra_item = not self._config.get_main_config().getboolean('show_heater_power', False)
```
lalten, line 189: JobStatusPanel inherits from ScreenPanel :)
```suggestion
extra_item = not self._show_heater_power
```

ks_includes/config.py
```diff
@@ -229,7 +229,8 @@ def _create_configurable_options(self, screen):
                 "range": [50, 150], "step": 1}},
             {"autoclose_popups": {"section": "main", "name": _("Auto-close notifications"), "type": "binary",
                                   "value": "True"}},
-
+            {"show_heater_power": {"section": "main", "name": _("Show Heater Power"), "type": "binary",
+                                   "value": "False"}},
```
lalten, line 233: nice!

lalten: Only thing that could be better still is that when that setting is changed KlipperScreen has to restart for it to become effective.

lalten:
Thanks for the changes @alfrix :) Tested this on my screen and it works great. Let's merge the PR?

alfrix:
the main issue with this currently is the merge of update_temp, they were different intentionally

multi-extruder setups broken, ks supports up to 5 extruders



also the degree symbol doesn't look good in the table and the percentage should be another column
or maybe just removing the column header will be better unsure

lalten:
Thanks a lot for your work @alfrix :1st_place_medal:

PR #713, opened 2022-08-29
Add Desktop file to easily start KlipperScreen on a normal Linux desktop

LordGaav:
Adds an example .desktop file, which when placed in `$HOME/.local/share/applications/` allows for easy starting of KlipperScreen.

Things that are nice to add:
* A KlipperScreen icon
* Instructions on how to install this (or integration in KlipperScreen-install.sh when a running `DISPLAY:0` is found?)


alfrix:
the file could be copied for all installs, it shouldn't be a problem
i can make an icon.

PR #717, opened 2022-09-03
Fixed Xclient scripts and added some examples

mattlalumiere:
xsdl script had incorrect script name (spelling error). Added some handling and daemonization of the service.

### DIFF
--- a/__main__.py
+++ b/__main__.py
@@ -1,0 +1,1 @@
+screen.py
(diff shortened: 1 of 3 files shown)


alfrix:
scripts/launch_klipperscreen.sh.EXAMPLE

that naming will lead to confusion adding 'sample-' at the start of the name would be preferred

PR #733, opened 2022-09-20
add .png icons support

pappicio:
Hi, with this pr, I ask to insert some code to enable importing of .png files also, so it will be easyer to make customs themes, iet tested and work, and also .png file support transaprent border, so, no problem about.
thanks.


pappicio:
add also .png (transparent borders) files into themes,  /images folder.

alfrix:
png is not a good idea for icons, icons should be resizable without quality loss, the ancient workaround was to have multiple sizes of the same icon to use depending on the resolution, now that we have vector graphics rasterized at runtime that hassle it's just not needed anymore

pappicio:
ok, you're right!
thanks!

Issue #748, opened 2022-10-05
Bl-touch does not raise z axis to deploy pin


kyledavis417:
**Bltouch can't deploy when z calibrate has been aborted and ran again**
If you run a z calibrate and abort and try to run the z calibrate again the probe can be too close to the bed and the pin can't deploy all the way causing you to have to home to the printer to try again.

**To Reproduce**
1. Go to z calibrate 
2. Click on start
4. move the nozzle down
5. Abort the process
6. go to z calibrate
7. click on start
8. pin hits the bed 
9. error in console that pin failed to deploy

**Expected behavior**
when you click start in z calibrate, if you have a bl touch the printer should rehome z axis before allowing the offset to be adjusted if the printer is homed.

alfrix:
this behavior is part of klipper, i don't think it should be addressed on the UI but in klipper itself

tinymachines3d:
@alfrix that's not the issue. Klipper Screen zcalibrate.py does not have a check for printhead location before starting z calibration. It only checks to see if the printer has been homed. If homed, it runs the script from wherever the printhead is. Say, for example, the user auto homes the machine, then homes only the Y axis and puts the printhead off the build plate (CORE XY), when you start Z calibrate, nozzle will crash.

alfrix:
zcalibrate does move to a configured or calculated position when doing probe_calibrate the other methods shouldn't require reposition
https://github.com/jordanruthe/KlipperScreen/blob/master/panels/zcalibrate.py#L148

tinymachines3d:
@alfrix 


If user aborts from Z Calibrate, uses actions>move, Z does not raise before moving to position. So, if the print head is close to the bed (after adjusting Z offset down, then clicking abort) bl touch fails to deploy because it hits the bed before any Z movement. 

Video link and klipperscreen log attached. 

https://youtu.be/2WYz3t5oMEM

[KlipperScreen.log](https://github.com/jordanruthe/KlipperScreen/files/9940991/KlipperScreen.log)

tinymachines3d:
@alfrix This only works with machines not using homing_override. For example if a machine needs to home Y before X, homing_override is required and safe_z_home has to be turned off. In this case a force Z raise would need to be ran before Z calibrate is ran again. The issue is the same as before [23b64b1](url) was pushed.

PR #751, opened 2022-10-06
Update zcalibrate.py

tinymachines3d:
Proposal: If abort is pressed, force auto home to reposition printhead when Z calibrate is restarted. 

Issue: In the case that the user aborts, the machine does not rehome. So, worst case, the printhead is off the bed on custom applications or in a place that the bltouch cannot probe, so when you press Start, it tries to home z in that position and moves below the gantry limits (corexy). Something will bend or break.

Currently, it checks to see if X,Y,Z are homed but doesn't check the current positions. Happy to see it fixed in a more elegant way but simplest thing was to remove the if statement.

### DIFF
--- a/panels/zcalibrate.py
+++ b/panels/zcalibrate.py
@@ -142,8 +142,7 @@ class ZCalibratePanel(ScreenPanel):
 
     def start_calibration(self, widget, method):
         self.labels['popover'].popdown()
-        if self._screen.printer.get_stat("toolhead", "homed_axes") != "xyz":
-            self._screen._ws.klippy.gcode_script(KlippyGcodes.HOME)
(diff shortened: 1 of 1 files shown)


tinymachines3d (commented):
Not sure why I changed then unchanged - ignore the first 2 commits. 

Proposal: If abort is pressed, force auto home to reposition printhead when Z calibrate is restarted. 

Issue: In the case that the user aborts, the machine does not rehome. So, worst case, the printhead is off the bed on custom applications or in a place that the bltouch cannot probe, so when you press Start, it tries to home z in that position and moves below the gantry limits (corexy). Something will bend or break.

alfrix:
this definitively not the way to solve #748

Issue #829, opened 2022-12-02
[BUG] KlipperSreen.desktop shortcut does not load on Gnome/Wayland


northcoastam:
**Describe the bug**
Trying to launch KlipperScreen as an app on Gnome with Wayland fails

screen.py fails at line 109: `monitor = Gdk.Display.get_default().get_primary_monitor()` because Wayland does not report a primary monitor.

Some detail here: No monitor reported as primary by GDK under Wayland https://gitlab.gnome.org/GNOME/gtk/-/issues/1028

**To Reproduce**
Steps to reproduce the behaviour:
1. Try to launch Klipperscreen.desktop on a desktop environment using Wayland.
2. Nothing happens
3. klipperscreen.log *may* show RuntimeError("Couldn't get default monitor").  I'm not sure why, but often the logging stopped just before the error.  See the attached log.

**Expected behavior**
KlipperScreen should start as expected
(I'm currently running Klipper headless on an old NUC and KlipperScreen on an Android phone, but trying to replace the two devices with an old touchscreen 2 in 1.  I installed Mint XFCE and then trying to get a better touch experience installed Gnome so ended up with Wayland.)

I added two lines to screen.py after line 109.  If there is no default monitor then it tries to get monitor 0;

```
109        monitor = Gdk.Display.get_default().get_primary_monitor()
           if monitor is None:
               monitor = Gdk.Display.get_default().get_monitor(0)
110        if monitor is None:
111            raise RuntimeError("Couldn't get default monitor")  

```
**Screenshots**

**Logfile**
The attached logfile shows a range of attempts to start KlipperScreen.desktop, mostly unsuccessful (although it would start in an xrdp remote desktop session) the last few a successful with the edits I made to screen.py
Attach your [klipperscreen.log](https://klipperscreen.readthedocs.io/en/latest/Troubleshooting/)
[KlipperScreen.log](https://github.com/jordanruthe/KlipperScreen/files/10137096/KlipperScreen.log)

PR #830, opened 2022-12-02
Possible fix for #829

northcoastam:
Fallback for Wayland when get_primary_monitor() is None.  Attempt to get_monitor(0) instead.


alfrix:
another possible fix would be always getting monitor 0 instead of the primary, but this is ok, thank you.

wayland also breaks the way ks setup dpms, but that shouldn't matter while being used as desktop app
and it doesn't set the window icon

northcoastam:
Thank you.  I’m busy googling what dpms is now 😝  I only noticed ks had its own icon a few days ago – I just took the generic one for granted, didn’t realise it was also a wayland thing.  I tried to change desktop environments but couldn’t – I think I’ve borked something with apt install.  Will probably wipe the touchscreen device and reinstall before migrating from the NUC/Android phone combo.

Thanks again for the sweet software.
Martin.

PR #861, opened 2022-12-26
bed_level: layout fix #891 improve screw detection close #895

### LINKED ISSUE #891: [BUG]
I have screws tilt adjust setup with 3 adjustment knobs that have been working properly for over 6 months. I'm assuming somewhere within recent updates a bug appeared within the screws tilt adjust menu of klipper screen. The icons seem too large for the screen now or just the bottoms cut off, if leave the menu all other menus now are missing the bottoms. The only way to clear is to change the theme, or restart klipperscreen from the web UI. This fault hasn't been noticed for me any other way UNTIL you go into the bed level, screws adjust menu. Then it appears and stays on all other menus until a reboot. 

The pictures show my default menu. And it all looks proper. If click into the screws tilt adjust menu with 3 bed screws( my scenario) the bottom is cut off. Not sure ifndoes it with 4 screws, I don't have that setup. I don't do a bed level very often until recently, so unsure if it was latest update or before.

From main screen go into configuration, 
then bed level. If screws tilt calc/adjust is setup I'm assuming, since that's my scenario.

### LINKED ISSUE #895: [FR] Improved bed_level.py for imperfect-grid screw layouts.
**Is your feature request related to a problem? Please describe.**

I have 8 screws on my modified ender 5 plus and those are not in a perfect grid.
This fails the simple bed_level.py screen button assignment algorithm. 

**Describe the solution you'd like**
I would like the buttons to be assigned properly as there is enough info to do so.
There is pull request implementing a more general fitting alg.
https://github.com/jordanruthe/KlipperScreen/pull/861

With the solution in the request KlipperScreen/bed-level.py is working for me 
as fabulous as other parts of this awesome project.

**Describe alternatives you've considered**
I have played with the options for KlipperScreen to no success.

**Additional context**
8-bed-screws

czaky:
First, many thanks for KlipperScreen. It is such an awesome and well designed UI. 

Added a better screw to button fitting procedure. This one tries to assign the buttons by eliminating the closest using a distance measure. It starts with the 4 maximal screws. Then it continues by trying the best fit to the middle of each of the 4 edges, for a total of 8 points max. This could be made more general, but I've tried not to disturb the code too much. Not sure if some of the remaining code, dealing with button association, is now obsolete. 

Added a `screw_button_scale` parameter. On  a 4.3" (800x480) screen, the UI would blow out with 8 screws assigned. Scaling it down with 2 instead of 2.5 allows all the buttons to fit on the screen.  Not sure if this should be an option or if the button_scale should be determined automaticaly.

Added a Z-Tilt before each adjust run, if the z-tilt was configured. Not sure if this should be an option as users can overwrite G28 and SCREWS_TILT_CALCULATE to Z-Tilt for them always.

Open for any suggestions.


czaky:
The find_closest function should probably take a max_distance parameter, to avoid assigning buttons across the printer-bed.

alfrix:
> Added a screw_button_scale parameter. On a 4.3" (800x480) screen, the UI would blow out with 8 screws assigned. Scaling it down with 2 instead of 2.5 allows all the buttons to fit on the screen. Not sure if this should be an option or if the button_scale should be determined automaticaly.

i don't like the button_scale configuration option, that should be resolved by the ui
but your current solution to #891 is acceptable now

i need to be test it in different screw configurations to see how it behaves

I'm sorry for the delay, I've been busy, thank you for not giving up.

alfrix:
lower the button_scale to 2 in all conditions 2.5 is only working for 4 screws, this must have happened when i did the window rezisability changes

alfrix:
Thank you for all your work it was merged in [55af0d7](https://github.com/jordanruthe/KlipperScreen/commit/55af0d753c2a9b0c9b606b069c55a6678ad67f6e), sorry for the delay

czaky:
Thank you for merging :)

PR #866, opened 2023-01-09
add ability to style and template menu buttons

geoffrey-young:


alfrix:
Thank you

thijstriemstra:
Thanks @geoffrey-young. Could you also include an example or two on https://klipperscreen.readthedocs.io/en/latest/Theming/? Not sure what src document it refers to though..

PR #870, opened 2023-01-18
Fix Bug - Changing wifi networks hangs

camerony:
Update wifi.py
Fix bug where switching wifi networks looks like this was introduced when this code was refactored and the {id} was not changed to {netid}


camerony:
More specifically this occurs when the previous wifi is not longer available

alfrix:
Thank you

PR #874, opened 2023-01-21
Update keyboard.py

camerony:
Update keyboard.py to support more valid characters for WiFi.  Additional pallet was created based on IOS WiFi keyboard.


alfrix:
Thank You

PR #960, opened 2023-04-26
Fixed z-bolt images opacity

JS-Cannon:
Fixed the opacity of a few images from the z-bolt style/images folder. Tested locally without any issues.


alfrix:
i don't see a reason for this, can you explain why this is needed?

JS-Cannon:
> i don't see a reason for this, can you explain why this is needed?

Greetings Alfredo. 

It is a mere aesthetic change. The commit takes care of those icons that are not fully transparent and have some dark shapes/pixels that show up albeit briefly when any user triggers a click/touch event.

If needed, I can post a link to a video showing this issue.

Also, I might be able to normalize the code further. AI breaks havoc on formatting.

alfrix:
i see it now, thank you

PR #978, opened 2023-05-15
Added and updated strings for DE | Added formal version for "Sie"

Xeroxxx:
- Added missing strings
- Adjusted strings to be more in line
- Same terminology for all strings
- _Added beside "Du" the formal version **de_formal** for "Sie"_


alfrix:
thank you

PR #989, opened 2023-05-24
Ukrainian localization update

komandrik:
Correcting the names of the previous translator and translating new meanings.


alfrix:
@alexeyzel can you comment because I see that retract has been changed back to the previous string, could you both agree to a translation?

alexeyzel:
I don't agree with translation of "Extrude" to "Заміна пластику" because it is not only menu name but also name of extrusion button. I propose to change it back to "Подача пластику". Everything else is OK with me.

alfrix:
> because it is not only menu name but also name of extrusion button

in English both are equal, so i don't consider this an issue
imagen
imagen


>  Everything else is OK with me

OK then i will merge this in the weekend

PR #1020, opened 2023-06-16
locale(pl): Update Polish translation

Daviteusz:
Some translations are bad and have not been tested by previous translators.
I will try to correct them and update the missing ones.

edit: Done. The translation is now almost perfect :D


alfrix:
thank you for your contribution

PR #1021, opened 2023-06-17
added steppers for vertical scrollbars fixes #1001 - moved to new PR

### LINKED ISSUE #1001: [FR] up/down arrows
Please add up and down arrows along the scrollbars, sometimes, when you have to scroll a little amount, would be easier to press the up or down botton twice than start dragging a scroll bar (that on some screen is small)

Vasilyev81:
ScrolledWindow has been replaced with ScrolledWindowWithSteppers in certain panels.

### DIFF
--- a/ks_includes/KlippyGtk.py
+++ b/ks_includes/KlippyGtk.py
@@ -251,3 +251,9 @@ class KlippyGtk:
                           Gdk.EventMask.BUTTON_RELEASE_MASK)
         scroll.set_kinetic_scrolling(True)
         return scroll
+
+    @staticmethod
+    def ScrolledWindowWithSteppers():
+        scroll = KlippyGtk.ScrolledWindow()
+        scroll.get_vscrollbar().get_style_context().add_class("with-steppers")
+        return scroll
--- a/panels/fan.py
+++ b/panels/fan.py
@@ -27,7 +27,7 @@ class FanPanel(ScreenPanel):
 
         self.load_fans()
 
-        scroll = self._gtk.ScrolledWindow()
+        scroll = self._gtk.ScrolledWindowWithSteppers()
         scroll.add(self.labels['devices'])
 
         self.content.add(scroll)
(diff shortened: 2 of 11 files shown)


alfrix:
hi, thank you for helping, a couple of notes:

always use a feature branch and not your master/main branch when doing PRs, so maintainers can push fixes if needed.

the increase in width is not something that i like here
you missed the retraction panel, mesh panel, exclude panel, etc
and i may put them behind an option in the settings

Vasilyev81:
WIP
Hi!
_"The increase in width is not something that i like here"_ - no problem, I'll return it back to original size, 2em is closest size I think
_"you missed the retraction panel, mesh panel, exclude panel, etc"_ - I did that way on purpose, I thought it is not necessary to have steppers in all panels, only in that panels where we have lists (of options for example) do you think it is necessary  to have on every panel? Then, probably I should also include steppers for horizontal scrolls?
_"and i may put them behind an option in the settings"_- you mean put it behind option like "show steppers"?

alfrix:
> I did that way on purpose,

ok, no problem, but the exclude panel is the one that may contain big lists, depending on how many objects are in the build plate, i placed 25 once  that's the advantage of exclude after all

>  "and i may put them behind an option in the settings"- you mean put it behind option like "show steppers"?

yes, maybe (not critical just a thought)

Vasilyev81:
OK, I hope I'm finally ready to make a pull request that doesn't need additional refinement. I've added steppers to retraction, mesh, and exclude panels. I would like to ask for some advice. Do I need to add steppes also to menu.py? What do you think? Also, I put steppers behind the option in settings. And is there a way to apply changes to this PR, or do I need to create a new one again?

alfrix:
>Do I need to add steppes also to menu.py

probably no,  you decide, i have two small 3.5" screens that i use regularly and i don't feel like they are needed anywhere.

> And is there a way to apply changes to this PR, or do I need to create a new one again?

you can edit the branch to merge at the top at the right of the tile, if you can't find it closing and creating a new one is ok too.

PR #1026, opened 2023-06-22
temp: add pid calibrate to the keypad

alfrix:


Guilouz:
Tested and it's working fine.

Maybe an improvement would be to add a confirm_send_action before starting pid.

PR #1029, opened 2023-06-25
Reorganization

alfrix:
imagen
imagen

### DIFF
--- a/ks_includes/defaults.conf
+++ b/ks_includes/defaults.conf
@@ -19,9 +19,56 @@ extruder = 210
 [menu __main]
 name: {{ gettext('Main Menu') }}
 
-[menu __main homing]
-name: {{ gettext('Homing') }}
+
+
+[menu __main move]
+name: {{ gettext('Move') }}
+icon: move
+panel: move
+
+[menu move homing homeall]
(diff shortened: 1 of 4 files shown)


Guilouz:
@alfrix Just try your new reorganization.

Do you think it's possible to adjust rows or colums to not have the slide bar and the icons at the bottom cut off ?

Best way would be to have all 4 lines displayed on screen.

My screen has a resolution of 1024x600 (like many screens).

IMG_0755

alfrix:
> Best way would be to have all 4 lines displayed on screen.

scaling can be tweaked to fit 4 rows, sadly the sacrifice will be icon size

> My screen has a resolution of 1024x600 (like many screens).

i only test 3 resolutions 480x320 800x480 and 1024x600, the only one i have to test physically is the small 480x320 3.5"

i have tried many other reorganizations, and this style of everything on the same place is the one i like the most

Guilouz:
@alfrix Yes new organization is good.
Yes reduce a little bit icons to have 4 rows will be good, to have up to 16 items by page.

Where i can tweak this to try and see result? I have try many things without success.

Guilouz:
@alfrix Confirmed it's working. Thanks.

PR #1031, opened 2023-06-30
added steppers for vertical scrollbars fixes #1001

### LINKED ISSUE #1001: [FR] up/down arrows

Vasilyev81:


Guilouz:
You have missed move and print menu no ?

Vasilyev81:
I don't see the reason to add it to the move and system because these panels have no lists.  I agree with you about the print panel.
About the speed - as far as I understand, there is no way to change the scrolling speed in GTK, it can be changed somewhere in os settings, but maybe I'm wrong. I would be grateful if you could tell me where to look for it.

Guilouz:
> I don't see the reason to add it to the move and system because these panels have no lists. I agree with you about the print panel. About the speed - as far as I understand, there is no way to change the scrolling speed in GTK, it can be changed somewhere in os settings, but maybe I'm wrong. I would be grateful if you could tell me where to look for it.

Move panel have scrolling when there is more stepper motors.

And as you can see system panel have scrolling bar : 

IMG_0762

Vasilyev81:
Well, you are right, I'll add it today. What about the speed? Do you know
how it could be adjusted in gtk3?

Guilouz:
@alfrix Possible to improve speed scrolling to this ? I have not found where.

alfrix:
> @alfrix Possible to improve speed scrolling to this ? I have not found where.

idk i will see if i can find something

alfrix:
the value in question is the step-increment
changing it with manually while running results in faster scrolling
imagen

the issue is that the value doesn't change, i tried with:
```
scroll.get_vadjustment().set_property("step-increment", 200.0)
scroll.get_vadjustment().set_step_increment(200.0)
```

making the step-increment equal to the page-increment would be ideal because the adjustment gets generated with values that depend on the content.

Vasilyev81:
Yeah, I tried that too, I mean to set up step increment property with different values. And the speed of the scroll hasn't changed.

guestisp:
Please merge this ...

PR #1071, opened 2023-08-05
Update Hungarian Language

AntoszHUN:
Quick update for the Hungarian language package.

  - Translation of newly added words.

Signed-off-by: AntoszHUN <zsikaiantal@gmail.com> or <zsikaiantal@icloud.com>


alfrix (changes requested):
This is not the correct location of the language

AntoszHUN:
> This is not the correct location of the language

Sorry. I put it in the wrong place.

PR #1090, opened 2023-08-20
UI tune fixes

blondak:


panels/spoolman.py
alfrix, line 1: This is not correct, the first param is the icon name, the issue must rely in klippygtk, i'll investigate

blondak: ok, then.

PR #1091, opened 2023-08-20
Updated ITALIAN Translation for Klipperscreen

pappicio:


pappicio:
Theese are italian translation files updated today, corrected some translations, fixed some italian words.

PR #1096, opened 2023-08-22
Spoolman: clear filter

blondak:


alfrix:
it updates the list correctly but it doesn't clear the selection dropdown.
also the dropdown starts blank and it should start with `All`

PR #1106, opened 2023-09-03
Led light control, close #991  

### LINKED ISSUE #991: [FR] LED
Hi, I'm Andrey.
From the beginning, I want you to know that I have no experience in contributing to an open source, so if I do something in an unusual/inappropriate way, do not hesitate to let me know about it, and please be patient.
In short, I have almost finished the development of the led panel, and just now I thought that (maybe) it is not implemented yet for some specific reasons.
The question is:  would you like to consider accepting a PR that implements led management functionality?
Below is a filled-out FR template from GitHub
**Is your feature request related to a problem? Please describe.**
It would be useful to be able to control the LED light from the KlipperScreen.

**Describe the solution you'd like**
I'm at the end of development and hope to finish it in the next few days. 

**Describe alternatives you've considered**
Currently, I have decided to go with scale for each color in rgb/rgbw mix; in the beginning, I considered using a color chooser widget, but I didn't find a way to customize it to RGBW color mix.

**Additional context**
Currently, as an environment, I have a Raspberry with a connected mouse and external screen and the second environment is  on Ender 3 V2 + Bigtreetech Manta E3EZ mainboard + Bigtreetech CB1 + sensor screen + neopixel led light.

Vasilyev81:
All functionality works for neopixel leds. It obviously needs some polish. Feel free to participate.
I would appreciate advice. I'll be a good Padawan.
Definitely needs:
~~title of current led color chooser with led name~~
~~add an icon with color to the main layout near the temp icons~~
-naming
~~not edited func show_error_modal in led~~


alfrix:
hi, apparently this has broken, i do not recommend doing merges from the main branch, it's much better to do rebase

if you give me an hour i can rebase and clean it

alfrix:
various issues here
the linter should be configured like:
https://github.com/KlipperScreen/KlipperScreen/blob/master/.github/workflows/test_code.yml
all the changes related to line length will be undone

seem that presets are taken from mainsail db, which is a big no, we cannot force people to use one webui over the other

the led icon could be something different but in the meantime light.svg was coded for this very usage

alfrix:
needs testing
changes:
removed the presets for now, needs some investigation
used light icon (already available for all themes)
removed the limitation of only 4 leds, it is unlimited now
removed the line spliting changes in al the other files
refactored the klippygcode to be more agnostic to the white color variable
refactored the color_mix getter to try find the available colors
created a function to open the led or color selector more cleanly
added colors for both presets selectors and leds selectors
probably other changes that i forgot.
TODO: 
the 4 bars color selection doesn't seem great to me, i will change it to a color picker
presets should have the color as an icon probably a colored circle.

alfrix:
idk how familiar are you with git but i suggest you do this to track the new changes:
```
git checkout led_control
git checkout -b led_control_OLD
git fetch
git checkout -b led_control origin/led_control
```

panels/led.py
```diff
@@ -0,0 +1,169 @@
...
+        grid = self._gtk.HomogeneousGrid()
+        color_data = self._printer.get_led_color(led)
+        color_mix = self._printer.get_led_color_mix(led)
+        if color_data is None or color_mix is None:
+            self.back()
+            return
+        if len(color_mix) == 3:
+            self.col_mix = RGB
+            if len(color_data) > 3:
+                # is it always rgbw?
```
Vasilyev81, line 86: From my experience, most led modules/led strips are 3 base colors [+ W warm/cold/light temp, K]

alfrix: it has nothing to do with the physical strips, the note was left as a reminder to check if klipper always uses 4 channels and the answer is yes
https://www.klipper3d.org/Status_Reference.html#led

Guilouz:
A good thing would be to be able to choose a pre-set for various events (like when starting klipperscreen, after printing, during bed/extruder heating, etc.).

Vasilyev81:
alfrix, about the presets from the mainsail. This feature was a killer from my point of view. Don't you want to rely on their APIs in this case?

alfrix:
> A good thing would be to be able to choose a pre-set for various events (like when starting klipperscreen, after printing, during bed/extruder heating, etc.).

that should be implemented with macros

> alfrix, about the presets from the mainsail. This feature was a killer from my point of view. Don't you want to rely on their APIs in this case?

you are not relying on an API, you are accessing their db directly, i can let it slide since it's not a professional project, but is not recommended as it can break at any point, the call needs to be done from the panel and it has to have more safeguards, you will see in the weekend when i give it a little more time.
and for presets fluidd needs to be supported too; both or none.

Vasilyev81:
I have working color buttons, but the led panel layout becomes a bit wide with these colored circles and it does not fit my small screen(

Vasilyev81:
>> and for presets fluidd needs to be supported too; both or none.
Of course, I understand you

alfrix:
Pedro told me presets are not supported in fluidd yet, so scratch that one

i've made a lot of mods to the panel:

* support for individual led channels not only rgb-rgbw
the config says:
>  The pin controlling the given LED color. 
>  At least one of the above parameters must be provided.
that means that support for any combination needs to be provided

* horizontal bars to avoid the visual glitch

* default presets are now always present, distributed buttons

* preview of the color

alfrix:
> move circles above/under the color label

above is the common place for the UI in this case

Vasilyev81:
If merge with current:

alfrix:
> Your opinion? Commit it?
`sure

Vasilyev81:
Before I pr, I have to fix one bug: the "back" function does not return to the "more" menu from the "Leds" choice menu

Vasilyev81:
Hi, I've tested it, but it be good if someone could re-test it. Also, I have this idea [(link to branch)](https://github.com/Vasilyev81/KlipperScreen/tree/btn_upd) but I think that the way I implement it is not the best from an architectural point of view. What do you think?

alfrix:
the main problem with the preview circles is that they break the label wrapping and ellipsizing, (try a really long label name)

Vasilyev81:
I thought about drawing this tiny colored stripe at the bottom of the button, then the length of the preset name is not important

alfrix:
> I thought about painting this tiny colored stripe at the bottom of the button with the required color, then the length of the preset name is not important

i actually ended up changing everything to previews

image

idk if that is better, what do you think?

Vasilyev81:
> idk if that is better, what do you think?
Looks nice to me.
Maybe swap circles with smth square-like?

alfrix:
image

> When is the next release date? What other features are planning to be included in oncoming release?

no ETA, and probably just this, i did not have a lot of time

i'll try one with labels to see how it looks

alfrix:
more than 5 chars and they break, i would prefer nothing like in mainsail
image
image

i will probably leave the color wheel or picker for later

Guilouz:
@alfrix, there is something wrong when there is many presets configured in mainsail, scrollbar encroaches on the icons

IMG_1287

PR #1121, opened 2023-09-16
Fixes for the Weblate repository

Daviteusz:
@alfrix  you can also download zip files with translations and reset the repository with this button 
obraz

There's also an add-on like this:
obraz
With it, you probably won't need to manually mark unused strings in po files.


alfrix:
i will refrain from doing anything with weblate until they approve the free hosting

Daviteusz:
So we are waiting. If they already approve, I will correct the PR if there will be any language changes.

PR #1143, opened 2023-10-10
Add nano editor commands for file edits

HumbleDeer:
This PR adds `sudo nano /path/to/file.ext` as well as closing/saving instructions

Adding nano edit commands to the page will improve people's ability to execute these steps. As noted during a Nero3D stream, neither our VzBot docs nor the page we linked to on KlipperScreen explained *how* to edit the file.

### DIFF
--- a/docs/Troubleshooting/Touch_issues.md
+++ b/docs/Troubleshooting/Touch_issues.md
@@ -65,15 +65,15 @@ DISPLAY=:0 xinput set-prop "<device name>" 'Coordinate Transformation Matrix' <m
 
 Where the matrix can be one of the following options:
 
- | Rotation  | Matrix |
- | - | - |
- | 0° | `1 0 0 0 1 0 0 0 1` |
- | 90° Clockwise | `0 -1 1 1 0 0 0 0 1` |
- | 90° Counter-Clockwise | `0 1 0 -1 0 1 0 0 1` |
- | 180° (Inverts X and Y) | `-1 0 1 0 -1 1 0 0 1` |
- | invert Y | `-1 0 1 1 1 0 0 0 1` |
- | invert X | `-1 0 1 0 1 0 0 0 1` |
- | expand to twice the size horizontally | `0.5 0 0 0 1 0 0 0 1` |
+| Rotation                              | Matrix                |
+| ------------------------------------- | --------------------- |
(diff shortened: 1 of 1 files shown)


alfrix:
Thank you, i will also clarify that the fix is only for raspberry

HumbleDeer:
Heads up or tip: Merging this with a squash merge would be beneficial, as it keeps your history tidy and uses feature names for your merges into main. :) 

My own names for commits really don't benefit being in the main branch by themselves; they're useful context in the PR but not after it's been merged.

HumbleDeer:
> Thank you, i will also clarify that the fix is only for raspberry

Can you elaborate on this or add some sources? I can have a quick go at adding another bit about this. I'm at my workbench anyway and don't mind doing this for you.

alfrix:
> > Thank you, i will also clarify that the fix is only for raspberry
> 
> Can you elaborate on this or add some sources? I can have a quick go at adding another bit about this. I'm at my workbench anyway and don't mind doing this for you.

/boot/config.txt it's not a standard linux file it only exists on RaspberryOS and derivatives (to my knowledge)

the fix is Raspberry specific because it's changing the driver (https://docs.mesa3d.org/drivers/vc4.html)
idk if other SBCs like the CB1 use the same gpu or not, but it's not a 'general' fix

HumbleDeer:
> > > Thank you, i will also clarify that the fix is only for raspberry
> > 
> > 
> > Can you elaborate on this or add some sources? I can have a quick go at adding another bit about this. I'm at my workbench anyway and don't mind doing this for you.
> 
> /boot/config.txt it's not a standard linux file it only exists on RaspberryOS and derivatives (to my knowledge)
> 
> the fix is Raspberry specific because it's changing the driver (https://docs.mesa3d.org/drivers/vc4.html) idk if other SBCs like the CB1 use the same gpu or not, but it's not a 'general' fix

Ah I see; That means the fix only applies to Raspberry Pi OS, but the issue could apply and need fixing on other platforms too. I have a CB1 based board (BTT Pi V1.2) but no screen to use it with.

alfrix:
the fix is for raspberrys because the issue is from raspberrys, other platforms would have other issues with different fixes that would need to be contributed from users (like the fix in question was)

PR #1146, opened 2023-10-15
Allow to use T0 even with 1 extruder

Otann:
I think it still a useful button, which can be used with a redefined T0 macro, which could be utilized with cooling down a hotend for example.

I had this on my printer before the update:

```
[gcode_macro T0]
description: When I press extruder button on the Extrude screnn in KlipperScreen
gcode:
  SET_HEATER_TEMPERATURE HEATER=extruder TARGET=0
```

### DIFF
--- a/panels/extrude.py
+++ b/panels/extrude.py
@@ -58,7 +58,7 @@ class Panel(ScreenPanel):
         i = 0
         for extruder in self._printer.get_tools():
             if self._printer.extrudercount > 1:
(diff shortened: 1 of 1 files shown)


alfrix:
this would need to be behind a check that T0 exists, because it was removed because people thought it was a bug.

and using T0 to turn off the heater its a terrible idea, some slicers like ideamaker output T0 in the Gcode by default and if your slicer of choice don't do it now, you can't be so sure it will not do it in the future.

Issue #1176, opened 2023-11-18
Option "width: ..." no longer works in v0.3.6


EmJay276:
### What happened?

By updating from v0.3.5-3 to v0.3.6-85 the following option is ignored and not working

```
[main]
width: 800
```

Here the relevant snipets of the log
0.3.5
```
2023-11-18 13:00:45,168 [screen.py:__init__()] - Screen resolution: 800x480
```

0.3.6
```
2023-11-18 13:02:23,925 [screen.py:__init__()] - Screen resolution: 848x480
```

I don't see any reason why the changes made to `screen.py` should cause this issue
https://github.com/KlipperScreen/KlipperScreen/compare/v0.3.5...v0.3.6#diff-4dc6710f3893ed193fc05caa325c9fe5b5d9e32e3bf2c73ae725a7dfbc1c9cd1
![grafik](https://github.com/KlipperScreen/KlipperScreen/assets/8619525/2361d2dc-2568-428c-8e23-1d3dd1ce53d3)

### What did you expect to happen instead?

Setting the screenwidth in [main] should still work

### How to reproduce this bug?

Update from v0.3.5-3 to v0.3.6-85

### Additional information:

I need to set this option due to weired config issue with my board/screen, see https://github.com/redrathnure/armbian-mkspi/issues/13 with the fix https://github.com/redrathnure/armbian-mkspi/issues/13#issuecomment-1633253747

### Log output
[logs-20231118-131711.zip](https://github.com/KlipperScreen/KlipperScreen/files/13400243/logs-20231118-131711.zip)

EmJay276:
I could work it down to the following commit breaking it https://github.com/KlipperScreen/KlipperScreen/commit/8f07f2197ebd9c9cbc5dc62f3268a9215bdbd6e8

51d21895d0599a645c51f4ac7785125c1bc792c0 is still working, 8f07f2197ebd9c9cbc5dc62f3268a9215bdbd6e8 is not working

EmJay276:
I found the issue causing this behaviour:
```
            self.set_resizable(True)
            self.windowed = True
```
is only set if it is a desktop screen, not if `self.width` or `self.height` are set anymore.
https://github.com/KlipperScreen/KlipperScreen/pull/1177 fixes this issue

EmJay276:
Fixed via https://github.com/KlipperScreen/KlipperScreen/pull/1177
@alfrix thank you very much! Probably best to remove the won't fix tag

PR #1177, opened 2023-11-18
Fix fullscreen issue when setting width or height in config for non-desktop machines

EmJay276:
Setting width and height in the config should result in a windowed screen, but it is only the case for desktop users. 
Fixes https://github.com/KlipperScreen/KlipperScreen/issues/1176 introduced via 8f07f21

### DIFF
--- a/screen.py
+++ b/screen.py
@@ -127,11 +127,13 @@ class KlipperScreen(Gtk.Window):
         self.width = self._config.get_main_config().getint("width", None)
         self.height = self._config.get_main_config().getint("height", None)
         if 'XDG_CURRENT_DESKTOP' in os.environ:
+            logging.warning("Running inside a desktop environment is not recommended")
             if not self.width:
                 self.width = max(int(monitor.get_geometry().width * .5), 480)
             if not self.height:
                 self.height = max(int(monitor.get_geometry().height * .5), 320)
         if self.width or self.height:
+            logging.info("Setting windowed mode")
             self.set_resizable(True)
             self.windowed = True
         else:
--- a/screen.py
+++ b/screen.py
@@ -131,6 +131,7 @@ class KlipperScreen(Gtk.Window):
                 self.width = max(int(monitor.get_geometry().width * .5), 480)
             if not self.height:
                 self.height = max(int(monitor.get_geometry().height * .5), 320)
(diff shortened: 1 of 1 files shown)


alfrix:
you are undoing the previous commit, i stand my ground no fullscreen under desktop environments and width is for development, fix your resolution issues

EmJay276:
@alfrix this won't change any behaviour introduced in https://github.com/KlipperScreen/KlipperScreen/commit/8f07f2197ebd9c9cbc5dc62f3268a9215bdbd6e8 to the desktop environment, neither revert it. Just fixes the bug introduced.

`if 'XDG_CURRENT_DESKTOP' in os.environ` it will set the `self.width` and `self.height` if not set in the config => `self.width` and `self.height` are always set after `L133` `if 'XDG_CURRENT_DESKTOP' in os.environ`
```
       if 'XDG_CURRENT_DESKTOP' in os.environ:
            if not self.width:
                self.width = max(int(monitor.get_geometry().width * .5), 480)
            if not self.height:
                self.height = max(int(monitor.get_geometry().height * .5), 320)
```

This lines here will result in exactly the same than in https://github.com/KlipperScreen/KlipperScreen/commit/8f07f2197ebd9c9cbc5dc62f3268a9215bdbd6e8, because `self.width` and `self.height` are set, except this will be also the case `if 'XDG_CURRENT_DESKTOP' NOT in os.environ` but `self.width` or `self.height` set in the config. => Same behaviour like 0.3.5 without the Bug introduced in https://github.com/KlipperScreen/KlipperScreen/commit/8f07f2197ebd9c9cbc5dc62f3268a9215bdbd6e8
```
        if self.width or self.height:
            self.set_resizable(True)
            self.windowed = True
```

EmJay276:
A different implementation with the same result would be

```python
        if 'XDG_CURRENT_DESKTOP' in os.environ or self.width or self.height:
            if not self.width:
                self.width = max(int(monitor.get_geometry().width * .5), 480)
            if not self.height:
                self.height = max(int(monitor.get_geometry().height * .5), 320)
```

But this will set the other one of `self.width` or `self.height` if only one was specified in the config.

alfrix:
ok, but you shouldn't be workarounding like this

PR #1211, opened 2023-12-21
add move_distances config option

### LINKED ISSUE #1210: [FR] Customize Move Distances
### What happened?

Currently the following values for the Move panel are [hardcoded](https://github.com/KlipperScreen/KlipperScreen/blob/master/panels/move.py#L11):

```python
distances = ['.1', '.5', '1', '5', '10', '25', '50']
distance = distances[-2]
```

My large printer uses something like:

```python
['.1', '.5', '1', '5', '100', '500', '1000']
```

Would be great to define these values in the configuration file instead. I'll see if I can come up with a pull request. For example:

```ini
# Define distances for the move panel
move_distances: 0.1, 0.5, 1, 5, 10, 25, 50
```

Ability to modify distances.

Go to move panel.

### Additional information:

_No response_

### Log output

```shell
● KlipperScreen.service - KlipperScreen
     Loaded: loaded (/etc/systemd/system/KlipperScreen.service; enabled; preset: enabled)
     Active: active (running) since Thu 2023-12-21 15:35:27 CET; 6min ago
   Main PID: 1182 (KlipperScreen-s)
      Tasks: 7 (limit: 765)
        CPU: 24.067s
     CGroup: /system.slice/KlipperScreen.service
             ├─1182 /bin/bash /home/thijs/KlipperScreen/scripts/KlipperScreen-start.sh
             ├─1185 /usr/bin/xinit /home/thijs/.KlipperScreen-env/bin/python /home/thijs/KlipperScreen/screen.py
             ├─1186 /usr/lib/xorg/Xorg :0
             └─1190 /home/thijs/.KlipperScreen-env/bin/python /home/thijs/KlipperScreen/screen.py

```
```

### PR DESCRIPTION
**Pull request by thijstriemstra**
fixes #1210

### CODE REVIEW

**Comment by thijstriemstra** [comment]
Unfortunately it's not working yet, maybe the config is `None` in the constructor? Could you take a look @alfrix?

File: panels/move.py [hunk 1]
```diff
@@ -13,6 +14,14 @@ class Panel(ScreenPanel):
 
     def __init__(self, screen, title):
         super().__init__(screen, title)
+
+        if self.ks_printer_cfg is not None:
+            dis = self.ks_printer_cfg.get("move_distances", '0.1, 0.5, 1, 50, 10, 25, 50')
+            if re.match(r'^[0-9,\s]+$', dis):
+                dis = [str(i.strip()) for i in dis.split(',')]
+                if 1 < len(dis) < 5:
```
alfrix, line 22: the array is checked to be less than 5 items and you passed 7

panels/move.py
```diff
@@ -13,6 +14,14 @@ class Panel(ScreenPanel):
 
     def __init__(self, screen, title):
         super().__init__(screen, title)
+
+        if self.ks_printer_cfg is not None:
+            dis = self.ks_printer_cfg.get("move_distances", '0.1, 0.5, 1, 50, 10, 25, 50')
+            if re.match(r'^[0-9,\s]+$', dis):
```
alfrix, line 20: the regular expression seems wrong, it doesn't work for floating point numbers (missing `\.` )

r'^[0-9,\.\s]+$'

panels/move.py
```diff
@@ -13,6 +14,14 @@ class Panel(ScreenPanel):
...
+                dis = [str(i.strip()) for i in dis.split(',')]
+                if 1 < len(dis) < 5:
+                    self.distances = dis
```
alfrix, line 23: missing default value reassign
 self.distance =  self.distances[-2]

alfrix (changes requested):

thijstriemstra:
thanks for the review, tested it and it works now.

alfrix (approved):

panels/move.py
```diff
@@ -13,6 +14,15 @@ class Panel(ScreenPanel):
 
     def __init__(self, screen, title):
         super().__init__(screen, title)
+
+        if self.ks_printer_cfg is not None:
+            dis = self.ks_printer_cfg.get("move_distances", '0.1, 0.5, 1, 50, 10, 25, 50')
+            if re.match(r'^[0-9,\.\s]+$', dis):
+                dis = [str(i.strip()) for i in dis.split(',')]
+                if len(dis) <= 7:
```
alfrix, line 22: if 1 < len(dis) <= 7:

alfrix (changes requested):

panels/move.py
alfrix: the list should have at least 2 items

thijstriemstra: > if 1 < len(dis) <= 7:

I tried initially but it didn't work. I'll try again..

> the list should have at least 2 items

ok

thijstriemstra:
Made the changes and all seems to work fine now.

alfrix (approved):

alfrix:
Thank you

PR #1218, opened 2023-12-28
Locale: Update sv-SE

eltacoloco:
Updated missing translations


eltacoloco:
Think weblate automaticaly merged my changes so din't ned this PR

PR #1219, opened 2023-12-28
feat: macro prompts close #1216

### LINKED ISSUE #835: [FR] confirmation to run macros
hello, 

it might be interesting to add a confirmation to run certain macros (for example a checkbox in action -> macro -> configuration), sometimes you get the wrong macro and you have to press the stop button urgency to stop it.

have a nice day.

### LINKED ISSUE #1216: [FR] Add support for macro prompt dialog
### Requested feature

Recently in mainsail (https://github.com/mainsail-crew/mainsail/pull/1630) and fluidd (https://github.com/fluidd-core/fluidd/pull/1262) a protocol has been implemented to support creating GUI prompts from macros.

The feature is documented here: https://docs.mainsail.xyz/overview/features/macro-prompts

### Is your feature request related to a problem?

For some macros like `SWITCH_FILAMENT` ([M600](https://marlinfw.org/docs/gcode/M600.html)), it is nice to be able to create a prompt for the user to for example purge more filament and resume the print. I imagine there are more use-cases that I have not thought about yet.

I think one big benefit of this is that less things have to be implemented in KlipperScreen and can instead be implemented through macros like https://github.com/KlipperScreen/KlipperScreen/issues/835, https://github.com/KlipperScreen/KlipperScreen/issues/89 or https://github.com/KlipperScreen/KlipperScreen/issues/56

### Describe alternatives you have considered

For the `SWITCH_FILAMENT` example, calling `PAUSE` and then overriding the `RESUME` macro works quite well with KlipperScreen, because it will display buttons for purging as well. The overriding of the `RESUME` macro is a bit more complicated, when you use macros like https://github.com/mainsail-crew/mainsail-config/blob/76a9999986b03abb3aed272dc0a818f13a18ea7e/client.cfg#L111-L117

I think introducing a different API just for KlipperScreen wouldn't make sense.

### Additional information

_No response_

alfrix:
now the user can have confirmation on macros using this feature close #835

limitations:
* button groups are not supported
* footer buttons and regular buttons are equal
* only 4 buttons are allowed (3 buttons are allowed in vertical/tate mode))

groups is technical limitation of the add_buttons method on gtk dialogs
this doesn't mean it's impossible, but a much more in depth and custom dialog not using the add_button method would be needed

imagen
imagen

### DIFF
--- a/ks_includes/KlippyGtk.py
+++ b/ks_includes/KlippyGtk.py
@@ -218,21 +218,30 @@ class KlippyGtk:
         if not self.screen.windowed:
             dialog.fullscreen()
 
+        max_buttons = 3 if self.screen.vertical_mode else 4
+        if len(buttons) > max_buttons:
+            buttons = buttons[:max_buttons]
+        button_hsize = max(len(buttons), 3)
         for button in buttons:
+            if 'style' in button:
+                style = button['style']
+            else:
+                style = 'dialog-default'
             dialog.add_button(button['name'], button['response'])
             button = dialog.get_widget_for_response(button['response'])
-            button.set_size_request((self.width - 30) / 3, self.height / 5)
-            format_label(button, 3)
+            button.set_size_request((self.width - 58) / button_hsize, self.height / 5)
+            button.get_style_context().add_class(style)
+            format_label(button, 2)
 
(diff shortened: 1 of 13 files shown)


alfrix:
this is how it would look like with 7 (apart of looking terrible it overflows)
imagen

PR #1246, opened 2024-01-17
Screen Brightness Panel + System Info Panel

rafaelSwi:
Two panels were added, which were added to the "More" menu (You can check their positioning in the defauls.conf file).

The **System Info** Panel (system_info.py) displays the host name, platform, version, system, and Mac address. To do this, I added an additional dependency, "get-mac".

The **Screen Brightness** Panel (brightness.py) is very simple and has just three vertically aligned buttons, "High", "Medium" and "Low", which regulates the screen brightness based on the user preference.

This has only been tested on a Raspberry Pi using Raspberry Pi OS, but if the procedure fails, or the necessary system file does not exist to change the screen brightness value, a message with "Error" will appear.

Also added ".vscode" and ".DS_Store" to gitignore.
".DS_Store" is the MacOS directory organization file, and ".vscode" is the junk that VSCode generates when coding.

I also applied a simple try-catch in config.py, in a code that defines the language to be used, on my mac, KlipperScreen crashes when starting because it doesn't know which language to set.

Some new words were also added, which were translated into all languages.
_(System Info, Platform, Version, Mac, Screen Brightness, High, Low)_

I hope it helped.


alfrix:
Hi, thank you
since i'm busy with other stuff it will take me a while to review all the code, it would have been easier if it was more split into different commits or PRs, but i'll get to it eventually

alfrix:
the gitignore lines have been merged

the issue with mac should be fixed with #1250 which makes the function cleaner

The info panel its a good idea, i don't think the mac address is something to show there, that should be in the network
in my opinion there is not enough info in there to justify a new panel, (it needs more info)

the brightness panel wouldn't work on a lot of systems, it needs to be present in the menu only if we are sure (at least to some extent) that it will work

f-strings cannot contain gettext _ functions because they will not be recognized by the update script, resulting in the removal of your translation lines when it's executed

Issue #1267, opened 2024-02-11
[BUG] The setting of the screws on the screen shows a checkmark icon even if the time difference is more than one hour


Pethical:
### What happened?

The setting of the screws on the screen shows a checkmark icon even if the time difference is more than one hour; however, the minutes are less than 6. For example, the checkmark appears even for 01:04.

### What did you expect to happen instead?

The checmark only show below **00**:06

### How to reproduce this bug?

1. Open Screws adjust menu
2. Rotate on of your screw to between 01:00 and 01:05
3. The checkmark will showing

### Log output

```shell
Not relevant
```

PR #1268, opened 2024-02-11
Fix bed level complete icon

Pethical:
This pull request fixes issue #1267.
The code was only checking the minutes, so when the hour was greater than 0, it still displayed the checkmark icon. This modification now also examines the hours.


alfrix:
Thank You

PR #1271, opened 2024-02-12
drop python 3.7 support

alfrix:


thijstriemstra (approved):

scripts/KlipperScreen-install.sh
```diff
@@ -98,7 +98,7 @@ check_requirements()
 {
     echo_text "Checking Python version"
     python3 --version
-    if ! python3 -c 'import sys; exit(1) if sys.version_info <= (3,7) else exit(0)'; then
+    if ! python3 -c 'import sys; exit(1) if sys.version_info <= (3,8) else exit(0)'; then
         echo_text 'Not supported'
```
thijstriemstra, line 102: Instead of `Not supported` print `Python 3.7 not supported`?

PR #1295, opened 2024-03-06
Improve params pattern to catch more macros

fbeauKmi:
Hello, 
The current pattern can only take into account params followed by default. The proposal catches more parameters and removes white spaces if present to avoid error sending macro
```elixir
[gcode_macro TEST]
gcode:
        {% set test = params.TEST_0 %}
        {% set test = params.TEST_1    |    default( 0 ) %}
        {% set test = params.TEST_2| int |default(0) %}
```


alfrix:
Good idea thank you

PR #1327, opened 2024-04-10
Tr_update

technical50:
April Turkish Language Update


alfrix:
you really need to use [weblate](https://hosted.weblate.org/engage/klipperscreen/) since there are a couple of translators for this lang and your changes will conflict with each other

Issue #1050, opened 2023-07-20
[FR] Beacon3D on Zcalibrate


stenmic:
hi, is it possible to add the BEACON_CALIBRATE command in the Zcalibrate panel. The Beacon3D does not work over Klipperscreen.

alfrix:
beacon_calibrate is not part of the klipper base, since it's an extra it cannot be part of the ui as a standard  that would have to be added as:
1. a macro  in the macros panel
2. as a button in your config

example of button:

```
[menu __main more beacon]
name: BEACON CALIBRATE
icon: bed-mesh
method: printer.gcode.script
params: {"script":"BEACON_CALIBRATE"}
confirm:
    {{ gettext('Run BEACON_CALIBRATE?') }}
```

PR #1350, opened 2024-05-02
Added beacon support

Zeanon:
As more and more people are running beacon, it would make sense to allow people to properly use KlipperScreen with that probe.
These changes just check whether the `beacon_calibrate` command exists and adjust accordingly. completely backwards compatible, no breaking change.

### DIFF
--- a/panels/zcalibrate.py
+++ b/panels/zcalibrate.py
@@ -62,6 +62,10 @@ class Panel(ScreenPanel):
 
         self.labels['popover'] = Gtk.Popover(position=Gtk.PositionType.BOTTOM)
 
+        self.probe_calibrate = ("BEACON_CALIBRATE"
+                                if "BEACON_CALIBRATE" in self._printer.available_commands
+                                else "PROBE_CALIBRATE")
+
         self.set_functions()
 
         distgrid = Gtk.Grid()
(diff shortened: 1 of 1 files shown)


alfrix:
As i answered in #1050 there are a lot of extra modules at this point and adding them would quickly result in a big burden, i would prefer sticking to mainline Klipper.

I've seen that fairly recently [probe_eddy_current](https://github.com/Klipper3d/klipper/commits/master/klippy/extras/probe_eddy_current.py) has been added to mainline

maybe the beacon can be configured as one of those? 

but it seems that work may not be complete: https://github.com/Klipper3d/klipper/pull/6558

Zeanon:
Doesn't seem like the best stance since beacon is becoming increasingly popular but can't be merged into mainline die to licensing issues but hey what do I know, I'm just a measily user wanting to properly use my printer.

alfrix:
>  I'm just a measily user wanting to properly use my printer.

that's how i started too

let's address the issues: it seems that you are making the `BEACON_CALIBRATE` a replacement for `PROBE_CALIBRATE` if it's available, that doesn't sound correct, it would be the same as doing:
```
[gcode_macro PROBE_CALIBRATE]
rename_existing: PROBE_CALIBRATE_BASE
gcode:
    BEACON_CALIBRATE
```

at that point i'm just confused on why they haven't registered the command as PROBE_CALIBRATE


in addition you need to add the section name here: (used to show the z_offset in various places)
https://github.com/KlipperScreen/KlipperScreen/blob/925dbe99aed1d1c1eb6073302ab930a93826ecd1/ks_includes/printer.py#L233

Zeanon:
that's a good point
Reason for not registering it as a probe is to avoid confusion, according to the beacon devs
Using the wrapper macro is a good idea though.
I might look into what needs to be modified as well to make it work properly, make it an additional method next to probe_calibrate and redo the pr I guess?

PR #1353, opened 2024-05-04
printer: tempstore fixup

linw1995:
The temperature retrieved from 'get_stat' could be {} if the device is unavailable, potentially resulting in a malfunction of the temperature panel.


alfrix:
Thank you

Issue #1384, opened 2024-05-30
[BUG] Syntax error in KlipperScreen-install.sh: line 154


NikolayMurha:
### What happened?

/home/pi/KlipperScreen/scripts/KlipperScreen-install.sh: line 154: syntax error near unexpected token `}'

### What did you expect to happen instead?

Installation

### How to reproduce this bug?

Try to install using KlipperScreen-install.sh

### Log output

```
/home/pi/KlipperScreen/scripts/KlipperScreen-install.sh: line 154: syntax error near unexpected token `}'
/home/pi/KlipperScreen/scripts/KlipperScreen-install.sh: line 154: `}'

#=======================================================#
 KlipperScreen installation failed!
#=======================================================#
```

HaniAlotaibi:
i just had the same issue exactly. my screen was not working (BTT HDMI5), removed the installed version, and when i tried to reinstall it, it  didnt work, direct install via KlipperScreen-install.sh and via Kiauh

dr-pavel:
I think we need to add the **fi** operator after 
these lines:
```
 if [ -d $KSENV ]; then
        echo_text "Removing old virtual environment"
        rm -rf ${KSENV}
```
You should get something like this:
```
if [ -d $KSENV ]; then
    echo_text "Removing old virtual environment"
    rm -rf ${KSENV}
    fi
```
then run the script:
`KlipperScreen-install.sh`

In my case, the installation was successful.

teunvandenburg:
> I think we need to add the **fi** operator after these lines:
> 
> ```
>  if [ -d $KSENV ]; then
>         echo_text "Removing old virtual environment"
>         rm -rf ${KSENV}
> ```
> 
> You should get something like this:
> 
> ```
> if [ -d $KSENV ]; then
>     echo_text "Removing old virtual environment"
>     rm -rf ${KSENV}
> fi
> ```
> 
> then run the script: `KlipperScreen-install.sh`
> 
> In my case, the installation was successful.

Works for me aswel thanks a lot.

Safiro1:
I having the exact same issue when trying manual installation or using KIAUH.
Please let us know when this is fixed.
Thanks!

Safiro1:
> > I think we need to add the **fi** operator after these lines:
> > ```
> >  if [ -d $KSENV ]; then
> >         echo_text "Removing old virtual environment"
> >         rm -rf ${KSENV}
> > ```
> > 
> > 
> >     
> >       
[...]

PR #1385, opened 2024-05-31
KlipperScreen Install Fix (Syntax error KlipperScreen-install.sh: line 154)

brunobold:
Fixed Syntax Error from line 154 causing install script to fail.


alfrix:
thanks, i'll reply in the bug thread

close #1384

PR #1404, opened 2024-06-14
startup script change sh to bash

breakheart:
'[[' used in line 14 is not supported by sh


alfrix:
i confirmed this issue when using wayland

alfrix:
thank you

Issue #1415, opened 2024-06-28
[BUG] Layer progress counter shows different numbers than Moonraker (or M117 ...)


kmarty:
### What happened?

In (Orca) slicer is set `SET_PRINT_STATS_INFO TOTAL_LAYER=[total_layer_count]` in "Machine start G-code" and `SET_PRINT_STATS_INFO CURRENT_LAYER={layer_num + 1}` in "Layer change G-code" (as well as `M117 layer {layer_num+1} / [total_layer_count]`), yet KlippeScreen shows different number of current layer than it should. Query against Moonraker API shows correct `current_layer`. The only place where KlipperScreen shows correct `current_layer` is in "Display message" line using `M117` G-code.

![20240628_165532](https://github.com/KlipperScreen/KlipperScreen/assets/2448725/f335546b-8cb3-4ade-a288-3d270844be3e)
![20240628_165538](https://github.com/KlipperScreen/KlipperScreen/assets/2448725/a7632634-b967-414f-a86c-36cc7e9afae8)

### What did you expect to happen instead?

To show correct `current_layer` (at least the same as `current_layer` in Moonraker API)

### How to reproduce this bug?

Just tap on Z-axis detail in KlipperScreen when printing _with any other layer height than 0.2mm_  (in this specific case it was 0.12mm layer height)

### Log output

[KlipperScreen-4bugreport.log.gz](https://github.com/user-attachments/files/16032601/KlipperScreen-4bugreport.log.gz)

kmarty:
I did some "investigation" and I believe, the issue is caused by this part:
```
...
            if 'info' in data["print_stats"]:
                if ('total_layer' in data['print_stats']['info']
                        and data["print_stats"]['info']['total_layer'] is not None):
                    self.labels['total_layers'].set_label(f"{data['print_stats']['info']['total_layer']}")
                if ('current_layer' in data['print_stats']['info']
                        and data['print_stats']['info']['current_layer'] is not None):
                    self.labels['layer'].set_label(
                        f"{data['print_stats']['info']['current_layer']} / "
                        f"{self.labels['total_layers'].get_text()}"
                    )
            elif "layer_height" in self.file_metadata and "object_height" in self.file_metadata:
                self.labels['layer'].set_label(
                    f"{1 + round((self.pos_z - self.f_layer_h) / self.layer_h)} / "
                    f"{self.labels['total_layers'].get_text()}"
                )
...
```
[...]

PR #1426, opened 2024-07-08
A crude way how to fix layer info inconsistency on job_status panel.

kmarty:
This is basically just attempt how to raise attention about issue #1415


alfrix:
i think it may be better to deprecate the calculation since it's likely to be wrong, if the user wants layer info they should provide it as it's documented, before the mechanism to set the current layer existed it did make sense to calculate an approx, but now it doesn't make sense anymore

Issue #1280, opened 2024-02-20
[FR] Tablet Support Meta Issue


ChaosBlades:
### Requested feature

This will serve as the meta issue for better tablet support in KlipperScreen.

I am creating this after seeing the Chris Riley video (https://youtu.be/QCAFqK_juQs?si=eC9eEBrqV5kbbvV4) and then realizing that old x86/ARM tablets used off eBay can be had for as little as $50 in excellent condition. I am currently testing functionality with a Microsoft Surface Go 2 I got for $72 running Ubuntu Server.

Will try to keep this updated with...
    - What features need to be added
    - What features are rejected and for what reason
    - Workarounds until features are added
    - General Progress or lack thereof

Current Limitations:
 - No way to tell what the battery percentage is
 - Auto Screen Brightness is not enabled by default
 - Screen Brightness defaults to 100%
 
Workarounds:
- View current battery percentage using command 'upower -i $(upower -e | grep BAT'. May not work in raspbian.

Feature Requests:
- Auto Detect when KlipperScreen is installed on a tablet by looking for a battery or having an option in KlipperScreen Setting to manually enable 'Tablet Mode'
- Battery Level indicator on Top Bar that also shows when the tablet is charging to the right of the clock https://github.com/KlipperScreen/KlipperScreen/issues/1262
- Menu in Settings called Tablet that shows the following options
    - Option to enable/disable battery percentage on Battery Level indicator
    - Option to adjust Screen Brightness
    - Option to turn on/off Auto Brightness
    - Option to enable/disable Screen Auto Rotation
    - Option to lock Screen to Landscape or Portrait
- Option in Klipper Settings for 'Extra Small' Font Size

PR #1437, opened 2024-07-20
Add battery percentage for tablets/laptops

### LINKED ISSUE #1262: [FR] Host system Battery Status
### Requested feature

Since i use a tablet and its battery powered i wanted to have a battery Status in the bar.

i know this Feture was requested ages ago and some users all over the web want such a feature.

i think about something:

make a configurable option like 
BATTERY_STATUS_COMMAND: cat /sys/class/power_supply/battery/capacity
which is send periodicly
BATTERY_STATUS_COMMAND_REFRESH: 30 #seconds to refresh

the Output has to be the status in percent, how the user get there is up to him, like with tee and/or some regex

also it would be possible to simply monitor a file on change so either the user writes a script wich periodically updates the file or such file exists like in my case:
/sys/class/power_supply/battery/capacity, under normal circumstances every battery powered linux device has such a file, the location only differs slightly, like someone has:
/sys/class/power_supply/surface-rt-battery/capacity.

Also possible to just run a gcode_shell_command macro and write the output to a file, but thats more complicated since normally klipper runs on another device than klipperscreen. at least in my case (but not impossible ssh will do just fine)

i hope i will knock on some open doors, in my case i would just programm it myself but im unable to write only a single liner in python. Away from this from what i know it cant be that hard to implement.

thanks in advance
phil2sat

### Is your feature request related to a problem?

no

### Describe alternatives you have considered

checked moonraker for such an option to just display it eg as fan but its not possible

### Additional information

_No response_

Tylerjet:
#1280  
closes #1262

Adds support for battery level status using psutil (currently polls every 30 seconds).
Indicators for Low Battery, Medium Battery, Full Battery, and Charging.
Togglable within KlipperScreen Settings.


alfrix:
* material icons do not look like any material icon i can source, they would need to be changed to proper material icons.
* are the icons used here free to use for opensource
* if there is no battery or not enabled, do not create the timer or call associated functions
* the tooltip should be more simple something like this:   "Show battery state if available"
* there is a lot of unnecesary repetition in that if, move `self.labels['battery_box'].pack_start(icon, False, False, 3)`
* you should split the function into at least 3 create, update and delete and call them where appropriate, calling a remove and recreating everything every 30 seconds seems very inefficient

panels/base_panel.py
```diff
@@ -191,6 +197,8 @@ def get_icon(self, device, img_size):
     def activate(self):
         if self.time_update is None:
             self.time_update = GLib.timeout_add_seconds(1, self.update_time)
+        if self.battery_update is None:
```
alfrix, line 200: only if the associated config is true

panels/base_panel.py
```diff
@@ -338,6 +346,47 @@ def update_time(self):
...
+
+        self.control['battery_box'].add(self.labels["battery_box"])
+        self.control['battery_box'].show_all()
+        battery = psutil.sensors_battery()
+        if battery:
+            battery_percent = int(battery.percent)  # Convert percentage to integer
+
+            if battery_percent < 33 and not battery.power_plugged:
+              icon = self._gtk.Image("low-battery", img_size, img_size)
+              self.labels['battery_box'].pack_start(icon, False, False, 3)
```
alfrix, line 369: unnecesary repetition
`self.labels['battery_box'].pack_start(icon, False, False, 3)`

panels/base_panel.py
```diff
@@ -338,6 +346,47 @@ def update_time(self):
...
+            elif battery_percent < 66 and not battery.power_plugged:
+              icon = self._gtk.Image("medium-battery", img_size, img_size)
+              self.labels['battery_box'].pack_start(icon, False, False, 3)
+            elif battery.power_plugged:
+              icon = self._gtk.Image("charging-battery", img_size, img_size)
+              self.labels['battery_box'].pack_start(icon, False, False, 3)
+            else:
+              icon = self._gtk.Image("full-battery", img_size, img_size)
```
alfrix, line 377: default should probably be an unknown icon ? add a full icon only if > 66

panels/base_panel.py
```diff
@@ -338,6 +346,47 @@ def update_time(self):
...
+              self.labels['battery_box'].pack_start(icon, False, False, 3)
+            
+            self.control['battery_box'].show_all()
+            battery_percent_str = str(battery_percent)[:2]  # Extract first two digits as string
+            print(f"Battery percentage: {battery_percent_str}%")
```
alfrix, line 382: don't use print use logger

panels/base_panel.py
```diff
@@ -338,6 +346,47 @@ def update_time(self):
...
+              self.labels['battery_box'].pack_start(icon, False, False, 3)
+            elif battery.power_plugged:
+              icon = self._gtk.Image("charging-battery", img_size, img_size)
+              self.labels['battery_box'].pack_start(icon, False, False, 3)
+            else:
+              icon = self._gtk.Image("full-battery", img_size, img_size)
+              self.labels['battery_box'].pack_start(icon, False, False, 3)
+            
+            self.control['battery_box'].show_all()
+            battery_percent_str = str(battery_percent)[:2]  # Extract first two digits as string
```
alfrix, line 381: do not create an addition variable, handle it in the f-string directly

panels/base_panel.py
```diff
@@ -338,6 +346,47 @@ def update_time(self):
             self.time_format = confopt
         return True
 
+    def battery_percentage(self, show=True):
```
alfrix, line 349: this is to big needs splitting and only updating if the ui elements are already in place, do not remove and recreate everything

alfrix (changes requested):

Tylerjet:
Hiya, first time with messing with python stuff so will look through.

Yes the icons are CC0 - Public Domain free to use, though with a quick google search i have found some material design ones that are open source and can use instead and may look neater. (These ones i just happen to find when messing with the solution late at night hehe.)

I should be able to resolve the other points as i go back through and make edits. Thanks for the review and comments!

panels/base_panel.py
```diff
@@ -117,6 +121,8 @@ def reload_icons(self):
             width = pixbuf.get_width()
             height = pixbuf.get_height()
             button.set_image(self._gtk.Image(name, width, height))
+        
+        self.battery_percentage()
```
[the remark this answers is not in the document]
Tylerjet: Would there be a better location to place this? or is it okay to leave as long as the config check is added?

The reason it was placed there is when the theme was changed the icon did not update if it was placed inside of ```def __init__```. Since that is not reinitialized after changing the theme as far as i can tell, so the next time the icon would update would be the next time the timer calls ```self.battery_percentage()```.

panels/base_panel.py
Tylerjet: I split out the creation and i think improved the updating but if it needs to be split more let me know.

panels/base_panel.py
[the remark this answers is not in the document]
alfrix: it should be called just before the timeout, think of it as: do it now, then every x time units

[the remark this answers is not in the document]
Tylerjet: Gotcha, yeah in the commits I pushed it now creates/inits the gtk box and labels in '''def __init__'''. Then in '''def reload_icons''' just calling '''battery_percentage()''' which just updates.

alfrix:
i don't have enough time now, but i will be merging this into a separate branch when i have more time to make the changes i want, you are free to continue improving it in the meantime

Tylerjet:
No Problem at all.

alfrix:
merged into a separate branch, you can target PRs to that branch instead of master until it's done, thank you

PR #1443, opened 2024-08-01
feat: brightness panel

alfrix:
first rough idea, not convinced this is the way forward, i want another solution

### DIFF
--- a/ks_includes/config.py
+++ b/ks_includes/config.py
@@ -166,6 +166,7 @@ class KlipperScreenConfig:
                 strs = (
                     'default_printer', 'language', 'print_sort_dir', 'theme', 'screen_blanking_printing', 'font_size',
                     'print_estimate_method', 'screen_blanking', "screen_on_devices", "screen_off_devices", 'print_view',
+                    'brightness_command'
                 )
                 numbers = (
                     'job_complete_timeout', 'job_error_timeout', 'move_speed_xy', 'move_speed_z',
--- a/panels/brightness.py
+++ b/panels/brightness.py
@@ -1,0 +1,74 @@
+import logging
+import subprocess
+import os
+import gi
+
+gi.require_version("Gtk", "3.0")
+from gi.repository import Gtk
+from ks_includes.screen_panel import ScreenPanel
+
+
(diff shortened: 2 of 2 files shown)


thijstriemstra:
I wonder if this'll work for screens i'm using: hyperpixel 4.0 and raspberry pi 7 inch official?

ChaosBlades:
@thijstriemstra Looks like if you have an entry under '/sys/class/backlight' then it should work. 

From my printers running KlipperScreen:
**Microsoft Surface Go 2** = `intel_backlight` | should work
**ZimaBoard + Waveshare 5.5in AMOLED** = no entry | Will not work | According to the Waveshare wiki you need the new version of the screen with updated firmware to control the backlight. Then you can adjust the backlight by holding the On/Off button
**Pi4B + Waveshare 4.3inch HDMI LCD (B)** = no entry | Will not work | According to the Waveshare wiki you need to modify the board by removing a resistor then solder on some wires from pads on the board to the PWM header on the Pi then you can control the brightness with gpio commands. Or just use the buttons on the side of the screen.

Your Screens:
**hyperpixel 4.0** = Will not work | Same situation as my 4.3" Waveshare | https://github.com/pimoroni/hyperpixel/issues/11
 **pi 7 inch official** = Should work | According to some internet searches this should have an entry in `/sys/class/backlight` called `rpi_backlight`

ChaosBlades:
@alfrix  If you are referring to calling/modifying `/sys/class/backlight` to adjust backlight as the way forward and not your implementation of it (I can't code) then yes that would be they way forward. It is the standard for how to control backlight from what I can see. If a display does not support this standard then it is going to be all over the place with how you adjust backlight. Would need to specifically support all these non-standard displays individually. As an example since I use Linux for all my PCs. I do not have an entry for my Desktop PC display even though KDE supports backlight control so that setting does nothing. HDMI 2.1 would need to work on AMD for me to get that brightness entry to show up and work. On my laptop I can control the backlight because I do have an entry. If you are talking about how you implemented it then I am sure there are plenty of open source examples from Linux desktop environments to compare with.

PR #1444, opened 2024-08-02
Update Ukrainian locale messages

mpyrozhok:
Some new messages do not have translation to Ukrainian. This commit adds the translation.


alfrix:
thank you

PR #1460, opened 2024-08-30
Make Z_ENDSTOP_CALIBRATE call BED_MESH_CLEAR only if bed_mesh enabled

smiksky:
When trying to calibrate the Z Endstop on a printer without mesh bed leveling, it throws an error popup because `BED_MESH_CLEAR` is not defined. This PR adds a check to only run that macro if `bed_mesh` is defined in Klipper.

I've tested the modification and it works as expected


alfrix:
Thank you

PR #1463, opened 2024-09-04
Add InputPupose DIGITS support to keyboard

CompyENG:
Adds support for InputPurpose.DIGITS to the on-screen keyboard.  This can be used in places where a number-only input is appropriate, but there may not be a desire to use the large keypad widgets.

A quick browse of KlipperScreen doesn't show anywhere useful to add this input purpose hint, but it could be used for custom panels.  The motivation for adding it comes from the Happy-Hare fork of KlipperScreen.  See also: https://github.com/moggieuk/KlipperScreen-Happy-Hare-Edition/pull/28

### DIFF
--- a/ks_includes/widgets/keyboard.py
+++ b/ks_includes/widgets/keyboard.py
@@ -23,7 +23,16 @@ class Keyboard(Gtk.Box):
         language = self.detect_language(screen._config.get_main_config().get("language", None))
         logging.info(f"Keyboard {language}")
 
-        if language == "de":
+        if self.entry_is_digital():
+            self.keys = [
+                [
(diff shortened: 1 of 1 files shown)


CompyENG:
Updated to fix style issues

alfrix:
i like this idea, but i do not want to merge code without any implementation.

i will make an effort to integrate it in the macros panel as an icon in the entry to select this version instead of the regular keyboard with a single touch, potentially useful for macros that require numerical input

i'll merge this when an implementation is ready to be merged together

alternatively, this could be integrated as an appended pallet to be accessible from the regular keyboard as an additional mode

PR #1464, opened 2024-09-04
Moonraker route prefix support

andrewmv:
Added optional `moonraker_path` config directive to allow connecting to a moonraker instance that uses a route prefix.
Added optional `moonraker_ssl` config directive to explicitly specify whether SSL/TLS should be enabled.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -49,8 +49,13 @@ Multiple printers can be defined
 [printer MyPrinter]
 # Define the moonraker host/port if different from 127.0.0.1 and 7125
 moonraker_host: 127.0.0.1
-# ports 443 and 7130 will use https/wss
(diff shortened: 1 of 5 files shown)


andrewmv:
Fixed an edge case that I thought might be helpful for other users, but I honestly don't possess the skills or patience to appease the code linter. If someone cares to offers guidance I'll do my best to apply it, otherwise feel to just close this.

alfrix:
Hi Andrew, i've fixed it, please do a git pull and verify that my refactorings are working correctly before the merge,
thank you

andrewmv:
Excellent, thanks for the help. Pulled this branch back down and it's still behaving correctly in my environment with every combination of the new options enabled/disabled.

PR #1485, opened 2024-10-16
camera: don't use scale filter if angle is zero

anarsoul:
scale filter seems to be expensive even with angle=0. On my CB1 having vf=rotate:0.0 in mpv arguments results in KlipperScreen using 100% of a single core vs 30% when it's omitted.


alfrix:
i haven't noticed any change in the Pi, but this couldn't hurt

Thank you

PR #1502, opened 2024-11-12
doc: add only_heaters

thijstriemstra:


alfrix:
In the docs i've stated:

> The options listed here are not editable from within the user interface.

There are many options that were removed from the docs, this is one of them, my position is that you should only configure things manually if there are no settings for it, otherwise you will have overrides that makes things not clear at all   when debugging

PR #1506, opened 2024-11-15
Fix screen not staying off

### LINKED ISSUE #1504: [BUG] Screen does not stay off
### What happened?

The screen turns off and immediately comes back on again, blanked, after `screen_blanking` timeout.
This repeats every `screen_blanking` seconds.

Screen stays off.

Enable screen blanking:
```
[main]
screen_blanking = 10
```

Wait.

### Additional information:

Introduced in f4c6801a8c562aa9f505864add76060d036b723b, previous versions work fine.

### Log output

[KlipperScreen.log](https://github.com/user-attachments/files/17757037/KlipperScreen.log)

apirek:
When the screensaver is active and the screen is turned off due to DPMS the unconditional call to self.screensaver.show wakes the screen up again.
The check whether the screensaver is already active got lost in the refactor.

Fixes #1504

### DIFF
--- a/ks_includes/widgets/screensaver.py
+++ b/ks_includes/widgets/screensaver.py
@@ -13,6 +13,9 @@ class ScreenSaver:
         self.screensaver_timeout = None
         self.blackbox = None
 
+    def is_showing(self):
+        return self.blackbox is not None
+
     def reset_timeout(self, *args):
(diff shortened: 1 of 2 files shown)


alfrix:
thank you

PR #1511, opened 2024-11-24
Spoolman Panel: Include Comments for Spools

Glought:
It simply adds the comment information from Spoolman under the spool's name, provided there is a comment associated with that spool.
KlipperScreen_SpoolMan_Comment_1

KlipperScreen_SpoolMan_Comment_2

This is useful if you have multiple printers and use the same brand ,type, and color of filament on each printer.

### DIFF
--- a/panels/spoolman.py
+++ b/panels/spoolman.py
@@ -335,6 +335,8 @@ class Panel(ScreenPanel):
             result = f'<big><b>{spool.name}</b></big>\n'
         else:
             result = f'<big>{spool.name}</big>\n'
+        if hasattr(spool, "comment"):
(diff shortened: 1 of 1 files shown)


alfrix:
before merging i need to test what happens if the comment is very long, will it wrap the lines or break everything

Glought:
I didn't think about that so I just tested it. It wraps the lines as seen in this Image.
KlipperScreen_SpoolMan_Comment_3

alfrix:
> I didn't think about that so I just tested it. It wraps the lines as seen in this Image. KlipperScreen_SpoolMan_Comment_3

thank you, this seems good enough

Issue #1325, opened 2024-04-08
[FR] Community themes


midikeyboard:
### Requested feature

I would think its a neat feature to add community themes similar to Mainsail and Fluidd.
this could be a page thats reads a simple table listing repo's made by github users. 

https://github.com/mainsail-crew/gb-docs/blob/main/_data/themes.csv
https://docs.mainsail.xyz/overview/features/themes/themes

ChaosBlades:
No time like the present.

My biggest issues with custom themes...
-  When a new icon is added it is simply missing. Would like it to at least revert to a default icon when a custom theme is missing that icon. 
- If KlipperScreen is re-installed the theme directory gets deleted but config still points to custom theme and KlipperScreen fails to start.
- If your KlipperScreen points to multiple printers you can't use a custom theme per printer.

![oled-dark](https://github.com/user-attachments/assets/335186f1-8dc8-4e05-ae87-e3b0306e8e6d)
[oled-dark.zip](https://github.com/user-attachments/files/17357134/oled-dark.zip)
![oled-darker](https://github.com/user-attachments/assets/44ab26c5-3d30-40c6-b98f-6c378f65e51f)
[oled-darker.zip](https://github.com/user-attachments/files/17357135/oled-darker.zip)
![purple-blue](https://github.com/user-attachments/assets/34d61d41-ec39-4c7d-949a-b3a862c28482)
[purple-blue.zip](https://github.com/user-attachments/files/17357136/purple-blue.zip)

PR #1536, opened 2025-01-24
Add new style: Modern-Ultradark

TheDayG0ne:
based on material-dark


alfrix:
i appreciate the effort, but it's a lot of effort to maintain a theme, i'm already regretting adding the extra ones that i made because  i have to make 5 different versions of each class everytime i want to add something

TheDayG0ne:
And then what should be done?

alfrix:
> And then what should be done?

probably a page in the docs like #1325 is asking for with screenshots and links

and a brief description on how to add it to the list

PR #1542, opened 2025-01-31
Fix heater graph not rendering by skipping data points that are None

jakpaul:
When there are data points in the temperature store that are None, the heater graph fails to render with an error like this:
```
2025-01-31 21:32:36,705 [functions.py:logging_exception_handler()] - Uncaught exception <class 'TypeError'>: '>' not supported between instances of 'NoneType' and 'float'
  File "/home/paulg/KlipperScreen/ks_includes/widgets/heatergraph.py", line 100, in draw_graph
    max_num = math.ceil(self.get_max_num(data_points) * 1.1 / 10) * 10

  File "/home/paulg/KlipperScreen/ks_includes/widgets/heatergraph.py", line 74, in get_max_num
    mnum.append(max(temp))
```
and
```
2025-01-31 21:34:01,713 [functions.py:logging_exception_handler()] - Uncaught exception <class 'TypeError'>: unsupported operand type(s) for *: 'NoneType' and 'float'
  File "/home/paulg/KlipperScreen/ks_includes/widgets/heatergraph.py", line 116, in draw_graph
    self.graph_data(

  File "/home/paulg/KlipperScreen/ks_includes/widgets/heatergraph.py", line 139, in graph_data
    p_y = max(gsize[0][1], min(gsize[1][1], gsize[1][1] - 1 - (d * hscale)))
```
This happened in my case after I added a temperature sensor to my printer. Since there were no data points before that, it defaulted to None, causing the error.
I would suggest fixing this by checking for None and skipping those points. See the changes

~ Paul

### DIFF
--- a/ks_includes/widgets/heatergraph.py
+++ b/ks_includes/widgets/heatergraph.py
@@ -132,7 +132,7 @@ class HeaterGraph(Gtk.DrawingArea):
         for i, d in enumerate(data):
             if d is None:
                 continue
-                
+
             p_x = i * swidth + gsize[0][0] if i != d_len else gsize[1][0] - 1
             if dashed:  # d between 0 and 1
                 p_y = gsize[1][1] - (d * (gsize[1][1] - gsize[0][1]))
--- a/ks_includes/widgets/heatergraph.py
+++ b/ks_includes/widgets/heatergraph.py
@@ -71,10 +71,10 @@ class HeaterGraph(Gtk.DrawingArea):
             if self.store[device]['show']:
                 temp = self.printer.get_temp_store(device, "temperatures", data_points)
                 if isinstance(temp, list):
-                    mnum.append(max(temp))
+                    mnum.append(max([v for v in temp if v is not None]))
                 target = self.printer.get_temp_store(device, "targets", data_points)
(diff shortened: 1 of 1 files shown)


alfrix:
i have never seen None as value, maybe this is new behavior. Did you ensure that the graph doesn't get distorted and looks like other interfaces?

jakpaul:
Somehow I am unable to reproduce the issue as it occured two weeks ago. This must be some edge case, might also be related to the toolchanger extension I am using.

I've done some more testing by inserting some dummy data, part of which is None, turns out with the current solution there is still the issue that filled graphs are not rendered properly (the dummy data is not None only where the line is):
2025-02-13-193735_1024x600_scrot
As you can see the final fill lines should not connect back to the chart bounds but to the lowest and highest x positions of the data points. I have added changes for that to my fork.
2025-02-13-193939_1024x600_scrot

There seems to be no distortion apart from this issue, both miniature and fullscreen graph look as expected. Though this fix does not cover the case when there is a gap in the data. It would then just connect across the gap. 
What do you think? Handling that will add some complexity for a case that probably rarely happens, although the added robustness would be nice I guess

PR #1580, opened 2025-05-18
Fix keyboard popup in Wayland/Cage

camelpunch:
Inside Cage (possibly any Wayland context?), the keyboard doesn't pop up. Strangely, the keypad does.

I tracked it down to `focus-in-event`, which doesn't appear to be triggered when touching an item such as the PSK password entry widget, whereas `touch-event` does.

I've tested this fork with X11 and the keyboard still pops up and submits the correct data, so it *appears* that this is safe, but I defer to your expertise as there may be other side-effects.


alfrix:
the problem here is that it doesn't pop-up for mouse interaction

camelpunch:
Ah I hadn’t realised that KlipperScreen supported mouse. Perhaps there’s a conditional approach, or a config option that could work then. I’ll have a closer look. Thanks!

PR #1609, opened 2025-08-12
Add caching for icons and pixbufs

as-com:
Profiling showed that a lot of time being spent repeatedly creating pixbufs for the same icons when loading panels. This change significantly improves responsiveness, especially on slower hosts.

Before:
speedscope before

After:
speedscope after


alfrix:
it does seems to be working correctly and it seems to be supported by python 3.2+ lgtm

PR #1633, opened 2025-12-30
Fix scrolling for vertical Spoolman panel

### LINKED ISSUE #1632: [BUG] Cannot use spoolman in vertical mode
### What happened?

When using klipperscreen in vertical mode, the scrollbar on the right overlays the checkboxes for spoolman spool selection, making it impossible to update/use spoolman.

the rounded corner sprite in image below is the scrollbar overlayed.

Image

that the scollbar doesn’t interfere with spoolman.

tested with 800x480 dsi in vertical mode (so 480x800)

### Additional information:

https://github.com/KlipperScreen/KlipperScreen/blob/0747a7a150a592be2b555d86b1f1aef6632cfec9/panels/spoolman.py#L187

### Log output

writing this on mobile

thijstriemstra:
fixes #1632

### DIFF
--- a/panels/spoolman.py
+++ b/panels/spoolman.py
@@ -315,7 +315,7 @@ class Panel(ScreenPanel):
         sort_btn_used.clicked()
 
         if self._screen.vertical_mode:
-            scrollbar_width = self.scroll.get_vscrollbar().get_preferred_width()[1]
+            scrollbar_width = 10
             self._treeview.set_margin_right(scrollbar_width)
 
         self.scroll.add(self._treeview)


alfrix:
the [revert](https://github.com/KlipperScreen/KlipperScreen/commit/b59b59becd586bffd7cc614e956065adb0b3deae) should suffice, i guess it needed overlay scrolling because the layout didn't adapt at that time, but i've tested and there is no need for it anymore, thank you and sorry for the inconvenience

thijstriemstra:
> the [revert](https://github.com/KlipperScreen/KlipperScreen/commit/b59b59becd586bffd7cc614e956065adb0b3deae) should suffice,

thanks, that also worked.

PR #1639, opened 2026-01-10
Exclude bad wifi interfaces

denywoo:
Exclude wifi interfaces in UNKNOWN, UNMANAGED or UNAVAILABLE states

KlipperScreen selects the first Wi-Fi interface returned by NetworkManager for connecting and scanning networks. If the system has more than one Wi-Fi interface and the first one is unavailable or unmanaged, this leads to a network scan error and makes connection management completely impossible. In my case, this was the p2p interface. I propose a simple solution that filters out interfaces in unsuitable states. This is not a definitive solution for choosing between multiple interfaces, but it helps avoid the most obvious errors.

### DIFF
--- a/ks_includes/sdbus_nm.py
+++ b/ks_includes/sdbus_nm.py
@@ -130,6 +130,11 @@ class SdbusNm:
             NetworkDeviceWireless(path)
             for path, device in devices.items()
             if device.device_type == enums.DeviceType.WIFI
+            and device.state not in {
+                enums.DeviceState.UNKNOWN,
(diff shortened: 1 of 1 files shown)


thijstriemstra:
>  i still don't have the necessary hw to test, i'll ask around if someone is willing to lend me a wifi adapter and maybe i can come up with something

(depending on the country you're in) I can send you a couple of different types of USB wifi dongles if you want. My email address is same as my git commit email if you want to get in touch @alfrix

PR #1649, opened 2026-02-13
feat: Add CARTOGRAPHER_SCAN_CALIBRATE to Z Calibrate

chris-red:
This change allows you to select CARTOGRAPHER_SCAN_CALIBRATE from the Z Calibrate Menu.


chris-red:
> it seems that CARTOGRAPHER_SCAN_CALIBRATE is not part of https://github.com/Klipper3d/klipper
> 
> commands not in official repo should use:
> 
> `zcalibrate_custom_commands`

Is this just a standard I should have used? Functionally it works as is.

Thanks,
Chris

chris-red:
Are I see you mean in the CONF, got it thanks

PR #1659, opened 2026-03-22
Add spoolman titlebar item with live updates

tetsu97:


alfrix:
i was just starting to review the code, something happened?

locki-cz:
We maked it together with tetsu and this was old version - not complete :( it was refactored and tested with small update, it works but maybe it is not clean enought to megre, i will make new PR soooon with small updates.

screenshot

locki-cz:
@alfrix crenated new PR, please check

PR #1663, opened 2026-04-04
Spool edit

locki-cz:
### This PR adds a dedicated spool editor

The spool editor writes spool weight changes directly to the Spoolman database, so values can be updated from KlipperScreen without going back to a computer

image

### What it adds:

- a new spool panel for editing spool weight directly in KlipperScreen
- support for opening either:
the Spoolman list (spoolman)
the active spool editor (spool)

- fallback to the Spoolman panel when no active spool is selected

### Spool editor behavior

- shows spool details and spool icon
- supports editing:
remaining weight
measured weight

- **Reset** fills in the maximum value for the currently selected mode
- updates are written back to Spoolman through Moonraker

- Tapping the spool description area or the spool icon toggles between Measured weight and Filament weight.

If you have any ideas for improving this, I’d be happy to hear them. I originally wanted to add a dedicated button below the spool info to switch between Measured weight and Filament weight, but I wasn’t able to make it work without breaking the layout, so for now it works this way.

It would also be possible to add a setting so it opens directly in the preferred mode, either Measured weight or Filament weight.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -98,13 +98,6 @@ moonraker_port: 7125
 # When the active spool drops below this value, the spool icon and weight turn red.
 # spool_low_limit: 20
 
-# Defines what the spool item in the title bar opens when tapped.
-# spoolman = Spoolman list
-# spool    = active spool editor
-# If not set, or if an invalid value is provided, the title-bar spool item
-# remains informational only and does not open a panel.
-# spool_shortcut: spoolman
-
 # Z probe calibrate position
 # By default it tries to guess the correct location
 # it will try using zero reference position, safe_z, mesh midddle, middle of axis length, etc
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -101,7 +101,8 @@ moonraker_port: 7125
 # Defines what the spool item in the title bar opens when tapped.
 # spoolman = Spoolman list
 # spool    = active spool editor
-# If not set, or if an invalid value is provided, the default is spoolman.
+# If not set, or if an invalid value is provided, the title-bar spool item
+# remains informational only and does not open a panel.
 # spool_shortcut: spoolman
 
 # Z probe calibrate position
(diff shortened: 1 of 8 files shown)


alfrix:
This is too much all at once, concentrate in the spool edit panel, and let's merge that alone first

thoughts on the the edit panel:

the title of the panel needs to be short and descriptive, in this case it would be something like "Spool Editor" or "Spool"

needs to be clear on what you are entering:
it needs either a label on top of the entry, or making the weights input entries like the spoolman page, so the right side panel is editing the current focused entry on the left

"filament weight" is not correct, it needs to be "remaining weight" or "used weight" depending on what you are reading

accessing trough a single click on the spool list adds a lot of area of potential misclicks, because of the usage of scroll by dragging the area instead of the scroll-anchor, an edit button will probably work better

the titlebar is not meant to be clickable because it's very slim and that makes it difficult to touch (most screens are in the 3.5-5" range), look at the size of the action bar or the scroll anchor for example, it's an interesting idea, but resizing the titlebar to be usable on click, will probably break most panels layouts, it's an interesting thought that can be explored later

locki-cz:
Yeah, it needs a bit of polish, thanks for the feedback! I'll fix that. I totally missed that I put 'spool name' instead of just 'spool' in the title bar – I guess I never really look at it! 

The spool menu also needs a quick rework, there’s too much dead space.

The shortcut is just a bonus. It works great on 4.3" and 5" screens, and if someone is on 3.5", they don't have to use it. ;)

locki-cz:
Fixed!

**Shortcut function removed, I will create separated PR later**

**Spool editor**

- the panel title no longer shows the filament name, the title is now always just Spool
- Filament weight was renamed to Remaining weight

**Spoolman**

- the spool name in the list is now shown as:
 Vendor - Name - Material
 for example: LDO - Black - ABS

- the separate Material column was removed
- it was replaced with a dedicated action column for opening the spool editor
- opening spool.py is no longer triggered by clicking the spool text or spool icon
- the editor is now opened only through a dedicated settings action icon

spoolman-new
spool-new

alfrix:
if anyone is wandering, this is on the spool-edit branch at the moment so it can be polished some more before merging

PR #1664, opened 2026-04-05
Refresh spool titlebar during prints - fix

locki-cz:
### What it changes:

- refreshes the active spool data every 10 seconds while printing or paused
- performs one final refresh after leaving the printing state
- forces a real reload from Moonraker/Spoolman instead of reusing cached spool data

### Why this is needed:

- before this change, the titlebar spool weight was only refreshed when the panel was opened or when the active spool changed during a print, the remaining spool weight could stay stale and would not update automatically
- after print completion, the displayed value could also remain outdated until another spool event happened

This keeps the titlebar spool weight current during the print and ensures it is refreshed once more when the print finishes.

### DIFF
--- a/panels/base_panel.py
+++ b/panels/base_panel.py
@@ -296,7 +296,7 @@ class BasePanel(ScreenPanel):
         if self.battery_update is None:
             self.battery_update = GLib.timeout_add_seconds(60, self.battery_percentage)
         if self.spoolman_update is None:
-            self.spoolman_update = GLib.timeout_add_seconds(10, self.update_spoolman_during_print)
+            self.spoolman_update = GLib.timeout_add_seconds(60, self.update_spoolman_during_print)
 
     def add_content(self, panel):
         printing = self._printer and self._printer.state in {"printing", "paused"}
--- a/panels/base_panel.py
+++ b/panels/base_panel.py
@@ -30,12 +30,14 @@ class BasePanel(ScreenPanel):
         self.time_format = self._config.get_main_config().getboolean("24htime", True)
         self.time_update = None
         self.battery_update = None
+        self.spoolman_update = None
         self.titlebar_items = []
         self.titlebar_name_type = None
         self.show_spoolman_in_title = False
         self.spoolman_low_limit = 20
         self.spoolman_icon_size = self._gtk.img_scale * self.bts * .9
         self.spoolman_icon_alert_pixbuf = None
+        self.spoolman_printing = False
         self.current_extruder = None
         self.last_usage_report = datetime.now()
         self.usage_report = 0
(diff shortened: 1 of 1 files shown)


alfrix:
wait, active polling every 10 seconds? :-1: 
 
i don't know how i'd missed that you don't have updates from moonraker

honestly i'm thinking that the feature needs to be reverted until we have a better  implementation

alfrix:
i'm putting it on a special branch for now

locki-cz:
**Fix fetches the data from Moonraker, not directly from the Spoolman database.**

**Specifically, it uses:**

- server/spoolman/spool_id
- server/spoolman/proxy with GET /v1/spool/{id}

**So the flow remains:**

- Klipper -> Moonraker
- Moonraker syncs data to Spoolman
- KlipperScreen queries Moonraker

**This is also the right approach from a performance perspective:**

- KlipperScreen does not need to handle a direct connection to Spoolman
- it uses the existing Moonraker integration layer
- it follows the same architecture as the rest of the integration

**The fix is lightweight because:**

- refresh runs only while printing or paused
- the interval is 10 seconds
- there is no periodic refresh at all when not printing
- after the print ends, it performs only one final refresh

**So in practice:**

- Moonraker may sync usage to Spoolman every 5 seconds
- KlipperScreen reads updated data only every 10 seconds
- this is lightweight and reasonable even for weaker hardware

**I think this is a good compromise:**

- we are not polling unnecessarily often
- we are not bypassing Moonraker
- and the UI still stays up to date during the print


This is default config, pushing update to spoolman server every 5 seconds, so we just ask moonraker every 10 seconds...
```
[spoolman]
server: http://192.168.101.12:7912
#   URL to the Spoolman instance. This parameter must be provided.
sync_rate: 5
#   The interval, in seconds, between sync requests with the
#   Spoolman server.  The default is 5.    
```

locki-cz:
> wait, active polling every 10 seconds? 👎
> 
> i don't know how i'd missed that you don't have updates from moonraker
> 
> honestly i'm thinking that the feature needs to be reverted until we have a better implementation

I is not active pooling, data about active spool are in moonraker already - it take data from moonraker, temperarute is updated more often.

locki-cz:
A user-configurable setting would likely resolve the issue: a 60s refresh rate for low-end machines and the default 5s for standard setups. However, it might not change much. We should merge the fix into the spool branch and get it tested by someone running on lower-end hardware.

alfrix:
i have been researching this yesterday, and verdict, at the moment the only alternative is to fetch on the moonraker notification about spool changed, and then use the filament used stat to calculate the difference until a new set_spool is emitted

not particularly keen on that approach either

because spoolman itself doesn't have object subscritption that i know off, lets go with polling, but lets do a sanity check for the time
for a density of 1.3g/cm3 (higher end of density for 3d filament) printing at 15mm3/s 
it would be 15mm3/second * 0.0013 g/mm3 = 0.0195 g/second = 1.17 g/minute
for a extreme end lets take 40mm3/s * 0.00125 g/mm3 (pla because petg doesnt flow that fast) we have 3g per minute

so the range is 20s to 60s to even register a change in the value assuming high speed printing and continuous flow, which is a pretty high bar, most printers that i've interacted with will not even register a change every minute

TLDR i think 60s is reasonable for this value

locki-cz:
> TLDR i think 60s is reasonable for this value

That sounds like a reasonable compromise; 60s is fine.

PR #1667, opened 2026-04-16
Fix spoolman refresh timer handling

locki-cz:
**Changes:**

- fix the set_spoolman_refresh() call signature mismatch
- keep the Spoolman refresh timer alive when not printing or while printer_select is open


alfrix:
completed with [405a462](https://github.com/KlipperScreen/KlipperScreen/commit/405a462ab1d440aa02a4afa08cebcf9f0db3ded6)

PR #1668, opened 2026-04-20
Fix double tap resume

chris-red:
Currently the resume button isn't deactivated after it is pressed, this means you can accidentally press it again while resuming, which can cause a move out of range and abort the print. This change disables the button after it is pressed in the same way pause is.

### DIFF
--- a/panels/job_status.py
+++ b/panels/job_status.py
@@ -421,8 +421,14 @@ class Panel(ScreenPanel):
 
     def resume(self, widget):
         self.disable_button("pause", "resume")
+        GLib.timeout_add_seconds(5, self._re_enable_resume_if_paused)
         self._screen._ws.klippy.print_resume()
         self._screen.show_all()
+
+    def _re_enable_resume_if_paused(self):
+        if self.state == "paused":
+            self.enable_button("resume")
+        return False
 
     def pause(self, widget):
         self.disable_button("pause", "resume")
--- a/panels/job_status.py
+++ b/panels/job_status.py
@@ -420,7 +420,6 @@ class Panel(ScreenPanel):
             logging.info(f"Could not restart {self.filename}")
 
     def resume(self, widget):
-        self.disable_button("pause", "resume")
         self._screen._send_action(widget, "printer.print.resume", {})
         self._screen.show_all()
 
--- a/panels/job_status.py
+++ b/panels/job_status.py
@@ -420,6 +420,7 @@ class Panel(ScreenPanel):
             logging.info(f"Could not restart {self.filename}")
 
     def resume(self, widget):
+        self.disable_button("pause", "resume")
         self._screen._ws.klippy.print_resume()
         self._screen.show_all()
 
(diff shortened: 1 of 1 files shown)


alfrix:
this would not work with macros that abort  the resume due to for example the extruder being not hot enough, look at [mainsail.cfg](https://github.com/mainsail-crew/mainsail-config/)

chris-red:
OK I'll look to add a timeout to it to restore the button after 5 seconds, sound good?

alfrix:
> OK I'll look to add a timeout to it to restore the button after 5 seconds, sound good?

try changing 
        self._screen._ws.klippy.print_pause()
into
        self._screen._send_action(widget, "printer.print.pause", {})

send_action has an integrated busy state that will wait until the websocket response, see if that is enough for your use case

chris-red:
> > OK I'll look to add a timeout to it to restore the button after 5 seconds, sound good?
> 
> try changing self._screen._ws.klippy.print_pause() into self._screen._send_action(widget, "printer.print.pause", {})
> 
> send_action has an integrated busy state that will wait until the websocket response, see if that is enough for your use case

I assume you meant changing the resume not pause right?

Trying that then hitting resume, changed the resume button to a pause one and resumed the print, hitting the pause button, while resuming appeared to send a pause which was eaten, giving the same type of issue you mentioned first but for Pause where the pause button was disabled while printing.

alfrix:
```python
diff --git a/panels/job_status.py b/panels/job_status.py
index d2113327..e9b1ef44 100644
--- a/panels/job_status.py
+++ b/panels/job_status.py
@@ -434,12 +434,12 @@ class Panel(ScreenPanel):
             self.restart_print()
 
     def resume(self, widget):
-        self._screen._ws.klippy.print_resume()
+        self._screen._send_action(widget, "printer.print.resume", {})
         self._screen.show_all()
 
     def pause(self, widget):
         self.disable_button("pause", "resume")
-        self._screen._ws.klippy.print_pause()
+        self._screen._send_action(widget, "printer.print.pause", {})
         self._screen.show_all()
 
     def cancel(self, widget):

```

chris-red:
> ```python
> diff --git a/panels/job_status.py b/panels/job_status.py
> index d2113327..e9b1ef44 100644
> --- a/panels/job_status.py
> +++ b/panels/job_status.py
> @@ -434,12 +434,12 @@ class Panel(ScreenPanel):
>              self.restart_print()
>  
>      def resume(self, widget):
> -        self._screen._ws.klippy.print_resume()
> +        self._screen._send_action(widget, "printer.print.resume", {})
>          self._screen.show_all()
>  
>      def pause(self, widget):
>          self.disable_button("pause", "resume")
> -        self._screen._ws.klippy.print_pause()
> +        self._screen._send_action(widget, "printer.print.pause", {})
>          self._screen.show_all()
>  
>      def cancel(self, widget):
> ```

AH OK got ya will try that thanks.

chris-red:
Works Great, thanks, have updated the PR.

PR #1684, opened 2026-05-06
feat: add support for hiding specific sensors via configuration

jLynx:
### Summary
This PR introduces a new configuration option, `hidden_sensors`, allowing users to explicitly hide specific sensors or devices from the UI. Previously, hiding was only supported by prefixing a device name with an underscore (`_`). This change enables more flexibility by allowing a comma-separated list of sensor names to be defined in the printer configuration.

### Changes
*   **Configuration Validation**: Added `hidden_sensors` to the validated strings list in `ks_includes/config.py`.
*   **Config Parsing**: Updated `base_panel.py` to parse the `hidden_sensors` string into a list of lowercase, stripped strings stored in `ScreenPanel`.
*   **UI Filtering**: Updated the `add_device` logic in both `main_menu.py` and `temperature.py` to check against the `hidden_sensors` list. If a device name (case-insensitive) matches a name in the list, it will not be displayed.

### Configuration Example
In your configuration file, you can now use:
```ini
[printer Printer]
hidden_sensors = Octopus_Pro_H723_V1.1,raspberry_pi
```

### DIFF
--- a/ks_includes/config.py
+++ b/ks_includes/config.py
@@ -190,7 +190,7 @@ class KlipperScreenConfig:
                 )
                 strs = (
                     'moonraker_api_key', 'moonraker_host', 'moonraker_path', 'titlebar_name_type',
-                    'screw_positions', 'power_devices', 'titlebar_items', 'z_babystep_values',
+                    'screw_positions', 'power_devices', 'titlebar_items', 'hidden_sensors', 'z_babystep_values',
                     'extrude_distances', 'extrude_speeds', 'move_distances', 'zcalibrate_custom_commands'
                 )
                 numbers = (
--- a/panels/base_panel.py
+++ b/panels/base_panel.py
@@ -543,9 +543,16 @@ class BasePanel(ScreenPanel):
                 logging.info(f"Titlebar name type: {self.titlebar_name_type} items: {self.titlebar_items}")
             else:
                 self.titlebar_items = []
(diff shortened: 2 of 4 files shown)


alfrix:
I'm struggling to see the value of adding this to the config file when the underscore convention already exists.

The only way this becomes a clear 'win' for the user is if there is a UI toggle (similar to how we handle the macros section) that allows users to hide/show sensors on the fly without editing files and restarting.

If a user has to open a config file anyway, it’s much simpler to just add an underscore to the name. What is the specific scenario where a user can't rename the sensor, or where a comma-separated list in the config is significantly better than the current naming convention?

jLynx:
@alfrix First off, thanks for taking the time to look at this PR and my other one. To answer your question, the main scenario is sensors defined in managed or included config files that the user doesn't control. In my case, RatOS defines [temperature_sensor Octopus_Pro_H723_V1.1] in its managed RatOS.cfg, and [temperature_sensor raspberry_pi] is also in an included config. These files are read-only in Mainsail/Fluidd. A user would have to SSH into the device to edit them, only for those changes to be overwritten on the next OS or framework update.

This applies to any setup where sensor definitions come from included configs maintained by a framework or plugin (RatOS, beacon, etc.) rather than the user's own printer.cfg.

I do agree a UI toggle would be the best experience - happy to look into that as a follow-up. But even with a UI toggle, the underlying mechanism would still need a way to persist the hidden state per-sensor, which is essentially what hidden_sensors provides.

PR #1685, opened 2026-05-06
feat: support dynamic registration of temperature sensors

jLynx:
### Summary
Some Klipper plugins (e.g., [Beacon](https://github.com/beacon3d/beacon_klipper)) create temperature_sensor objects dynamically at runtime rather than through printer.cfg. These sensors don't appear in Klipper's configfile API response, so KlipperScreen never registers them, never subscribes to their updates, and they're completely missing from the UI, even though they're visible in Mainsail/Fluidd.

This PR fixes that by querying Klipper's object list during initialization to discover and register any temperature sensors that aren't in the config.

image

### Changes
*   **Dynamic Sensor Registration**: Added `register_dynamic_sensors` to `ks_includes/printer.py`. This method scans the Klipper object list for `temperature_sensor` entries, filters out hidden ones (starting with `_`), and initializes them in the internal data structure if they are missing.
*   **Startup Discovery**: Updated `screen.py` to fetch the full list of printer objects from the Moonraker API during Klipper initialization and pass them to the registration logic.
*   **Reactive UI Updates**: Updated `process_update` in both `main_menu.py` and `temperature.py` to check if a device exists in the current panel. If a sensor update is received for a known printer device that hasn't been added to the UI yet, it is now dynamically added to the panel.

### Technical Details
- Increments `tempdevcount` for each newly discovered sensor to ensure accurate device counts.
- Maintains existing naming conventions (respecting the `_` prefix for hidden sensors).
- Ensures that sensors appearing mid-session (via `process_update`) trigger the `add_device` UI logic to prevent missing entries on the display panels.

### DIFF
--- a/ks_includes/printer.py
+++ b/ks_includes/printer.py
@@ -101,6 +101,18 @@ class Printer:
         self.log_counts(printer_info)
         self.process_update(data)
 
+    def register_dynamic_sensors(self, objects):
+        for obj in objects:
+            if obj.startswith("temperature_sensor "):
+                name = obj.split(" ", 1)[1]
+                if name.startswith("_"):
+                    continue
+                if obj not in self.config:
+                    logging.info(f"Registering dynamic sensor: {obj}")
+                    self.config[obj] = {}
+                    self.data[obj] = {"temperature": 0}
+                    self.tempdevcount += 1
+
     def log_counts(self, printer_info):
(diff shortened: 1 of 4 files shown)


alfrix:
LGTM

alfrix:
thank you for contributing

PR #1686, opened 2026-05-11
install: print system is going to reboot

thijstriemstra:
It's always a bit surprising to see the klipperscreen install end with:

```
Summary:
  Upgrading: 0, Installing: 0, Removing: 0, Not Upgrading: 0
Starting service...
Connection to myprinter.local closed by remote host.
Connection to myprinter.local closed.

```

and suddenly close ssh and reboot the system. Added a message to let the user know what's going on.


alfrix:
you are right, thank you

PR #1688, opened 2026-05-18
Add Serbian and Croatian Latin translations for KlipperScreen

InfinitumForm:
Adds native Serbian (Latin) and Croatian Latin translations for KlipperScreen to improve accessibility and user experience for Balkan communities.


thijstriemstra:
hvala!

alfrix:
Thank you for contributing!

InfinitumForm:
My pleasure, and thank you as well for maintaining such a great project. I've been using KlipperScreen for quite some time on my custom 3D printer setup and I'm genuinely very satisfied with it. I may have a few ideas and improvement suggestions in the future, but I'll share those another time. Thanks again for the merge and for all the work behind the project. 🙂

thijstriemstra:
maybe good to add this to the changelog, so those speaking these languages will notice and can switch away from english if they want.

InfinitumForm:
> maybe good to add this to the changelog, so those speaking these languages will notice and can switch away from english if they want.

That's actually a very good point, thank you. I wasn't sure about the project's changelog workflow, so I preferred not to touch it. But I completely agree that it would help users notice the new language support more easily.

PR #1694, opened 2026-05-26
USB File Copy Functionality & Documentation

TheyCallMeZ:
- Add usb_drives.py for USB Menu
- Update base_panel.py to show the menu option in the left panel 
- Update KlipperScreen-install.sh to facilitate the auto USB drive mounting 
- Add usb-pen-drive-icon.svg from https://uxwing.com/usb-pen-drive-icon/ (License states without attribution) 
- Update Documentation to show USB Drive Copy Panel
- Add USB Copy Documentation showing workflow.


alfrix:
Thanks for the PR, but i have a few concerns about the current approach.

KlipperScreen is designed to be a client frontend. people frequently run it on remote, non-host devices.
A frontend installing automount breaks that separation of concerns.
If the host or Moonraker isn't handling the mounting and exposing the files on the backend, the client shouldn't be forcing it from the front.
I don't think we can merge this implementation as-is, because i think that this functionality belongs inside the G-codes panel, showing up as a directory or source instead.

TheyCallMeZ:
> Thanks for the PR, but i have a few concerns about the current approach.
> 
> KlipperScreen is designed to be a client frontend. people frequently run it on remote, non-host devices. A frontend installing automount breaks that separation of concerns. If the host or Moonraker isn't handling the mounting and exposing the files on the backend, the client shouldn't be forcing it from the front. I don't think we can merge this implementation as-is, because i think that this functionality belongs inside the G-codes panel, showing up as a directory or source instead.

Hi alfrix! 

I understand your concern, the use case of this implementation is targeted at those who either travel with their printers, or run their machines air-gapped post-set up which I believe is the default installation and behavior of KlipperScreen. Having the option to move files within the gcodes section felt as it would be adding steps into an existing workflow and potentially causing frustration to the user when they wanted to just print and are now presented with extra options of where to print from, or what the workflow would be for a file copy operation. 

With regard to the auto-mount, and enabling it during installation, I can concede in a remote non-host device situation it is less than ideal, and could be segregated to it's own post-install script. Potentially having an environment variable check before showing the icon in the left panel. For what it's worth, in the communities I have been a part of KlipperScreen running  on the host is the dominant use case and was the impetus for making the feature in this way.  

I would greatly like to see this feature implemented, and would ask for your consideration if it were modified to be an option to enable rather than part of the default install.

Thanks again for taking the time to review this PR!

alfrix:
the community seems to recommend moonraker-loader for this usecase
https://github.com/shiftingtech/Moonraker-loader/

it combines automount with a script that instead of copying, it will pipe it through curl to moonraker, that way the metadata parsers and post-process are run on the file

you could in theory automount the usb to a folder inside the gcodes root, but you will not have the aforementioned parsing, and on top of that printing from the usb is discouraged due to poor performance from the drives that could cause timer too close errors

TheyCallMeZ:
> the community seems to recommend moonraker-loader for this usecase https://github.com/shiftingtech/Moonraker-loader/
> 
> it combines automount with a script that instead of copying, it will pipe it through curl to moonraker, that way the metadata parsers and post-process are run on the file
> 
> you could in theory automount the usb to a folder inside the gcodes root, but you will not have the aforementioned parsing, and on top of that printing from the usb is discouraged due to poor performance from the drives that could cause timer too close errors

Thanks for the info! I took a look at moonraker-loader and it's definitely a one-way solution with no user feedback for when the operation is complete either. Not having the option to print from the USB drive is a great callout and not something that I had thought about. It is definitely going to have to be on my list of scenarios to potentially mitigate. I have opened a PR with the moonraker team and provided some additional information to them on the PR. 

I have found that if I automount into the gcodes directory, I would be able to access the files and copy via Moonraker. So the primary things that are needed are listing the currently mounted USB Drives and being able to eject drives via Moonraker. The alternative would be to add a new mounts or usb folder to the array of folders Moonraker allows to select from. 

Short Version: Stay tuned! I am looking to see about getting this going on the moonraker end, and that will most likely determine how I change the UI in this PR.

alfrix:
i've forked it: https://github.com/alfrix/Moonraker-loader/

added notifications through RESPOND klipper module that should appear on all UIs
moves the uploaded gcodes from the usb to an uploaded folder in the usb

no changes are needed in KlipperScreen, the solution is universal

alfrix:
the closure is for this specific proposal, you can open a new one whenever you like

PR #1704, opened 2026-06-14
Add config-driven keybinding system for keyboard navigation

enricolam:
## Summary
  Adds a config-driven keybinding system for keyboard navigation and alternative input devices where key mapping might not be an option.

  ## Changes
  - Config-driven keybinding system with keygroups, accumulators, and context resolution
  - Panel-specific and global keybinding support
  - Removed hardcoded Escape/Backspace (now opt-in via config)
  - Security: accumulators cleared on screen lock
  - Various standard US ASCII keygroup avaliable as library
  - Example configurations for basic and advanced usage
  - Complete documentation

  ## Use Cases
  - Custom accessibility devices with uncommon keys
  - HID device integration (barcode/QR scanners)
  - General keyboard navigation

  ## Testing
  Tested on local machine (Ubuntu 22.04 LTS paired with mainsail-crew/virtual-klipper-printer)

  ## Breaking Changes
  None. Keybindings are opt-in; existing behavior re-implemented as config (default_navigation.cfg).

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 KlipperScreen is a touchscreen GUI that interfaces with [Klipper](https://github.com/Klipper3d/klipper) via [Moonraker](https://github.com/arksine/moonraker). It allows you to switch between multiple printers and access them from a single location. Notably, it doesn't need to run on the same host as your printer; you can install it on another device and configure the IP address to connect to the printer.
 
+KlipperScreen also supports keyboard navigation and customizable shortcuts, enabling keyboard-only operation and alternative input devices like barcode scanners and arcade controls.
+
 ### Documentation
 
 For detailed information, [click here to access the documentation](https://klipperscreen.github.io/KlipperScreen/).
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -261,6 +261,60 @@ icon: heat-up
 panel: preheat
 ```
 
+## Keybindings
+
+Keybindings provide keyboard navigation and customizable shortcuts. See the [Keybindings documentation](Keybindings.md) for complete details.
+
+### Quick Start
+
+```{ .ini }
+# Enable default Escape/Backspace behavior
+[include ~/KlipperScreen/keybindings/default_navigation.cfg]
+```
+
+### Section Types
+
+The keybinding system uses four configuration section types:
+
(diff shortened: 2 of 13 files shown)


alfrix:
It's an interesting idea but i have a major concern:
The accumulator implementation is too dangerous as-is. It turns KlipperScreen into an unvalidated gcode terminal, which is a security risk for a device that may be shared or accessible to non-technical users.
Accumulators let raw keystrokes become gcode commands and execute them. There is zero validation of what gets accumulated
The confirm dialog isn't enough, it's optional (not required), doesn't prevent typing dangerous commands, and doesn't prevent accidental execution.

barcode/QR scanner use case:
the safer approach is creating a panel for the reading, that parses the input properly.

Custom accessibility devices:
you can map keys with tools like keyd (I use it for the copilot key).

and there are very questionable decisions:
in the example "basic" like mapping F1 which is traditionally the help key as an E-stop
it disables the current keybindings by default

Minor issues
* test_config.cfg is referenced in the docs but doesn't exist in the PR.
* get_active_bindings silently overrides, at least log a collision warning.

While I appreciate the effort that went into implementing this complex system, it introduces too many safety risks and configuration complexities to be maintainable. The correct approach for scanners is a dedicated panel that focuses strictly on sanitized barcode/QR input. Again, I appreciate the time and effort you put into the concept, it's just too complex for KlipperScreen and belongs in a standalone external input parser.

enricolam:
Thanks for your time for reviewing the commits, honestly the code is designed for a very different use case in mind, my printers and machines on Klipper are primarily for a broader audience in the University, where you can think of it similar to a high tech vending machine, using keyd is definitely feasible, but becomes rather difficult to synchronize the behaviour between a dozen of machines, the QR part is sanitized by another piece of software which does not present (I have tested that most malformed gcode are refused by Klipper) such that safety is considered (it's not really a machine, it's a very interesting teaching rig that does something else than anykind of machining).

I totally get where your concerns are from tho, this feature is designed for team collaboration with extensive training and supervision, an assumption where they should know what is safe and what is not, perhaps for the use-case of this repository adding a mandatory list buried in the python is more appropriate, anyway, really thanks for your time for reviewing these spontaneous commits 🥂

PR #1728, opened 2026-07-11
Increase precision of pressure advance label

thijstriemstra:
While printing the [Pressure Advance calibration test](https://www.orcaslicer.com/wiki/calibration/pressure_advance_calib#line-method) in Orcaslicer, I noticed the PA value only changed every 0.1 increment, not showing the actual pressure advance increase of 0.02:

image

Turns out the decimals was set to 2, so I increased it to 3.


alfrix:
Makes sense for direct extruders, thank you

PR #1729, opened 2026-07-12
Cache compiled Jinja templates used by menu enable evaluation

JP-Reitsma:
## Summary

Cache compiled Jinja templates used by `evaluate_enable()`.

Previously every call parsed and compiled the template before rendering it, even when the same enable expression was evaluated repeatedly.

This change caches compiled templates keyed by their template string and reuses them on subsequent evaluations.

## Benefits

- Avoids repeated Jinja compilation
- Preserves existing behaviour
- Dynamic expressions are still rendered on every evaluation
- Improves menu responsiveness on lower-powered hardware

## Notes

Special cases (`True`, `False`, and `{{ moonraker_connected }}`) continue to use the existing fast path.

### DIFF
--- a/panels/menu.py
+++ b/panels/menu.py
@@ -5,7 +5,6 @@ import gi
 
 gi.require_version("Gtk", "3.0")
 from gi.repository import Gtk
-from jinja2 import Template
 
 from ks_includes.screen_panel import ScreenPanel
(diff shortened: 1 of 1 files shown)


alfrix:
Thank you for the PR, at first glace it does look good, i'll test it a bit before merging

PR #1730, opened 2026-07-13
Avoid unnecessary reloads of the G-code macro panel

JP-Reitsma:
This change avoids rebuilding the G-code macro panel when it is revisited.

The macro list is now populated only when needed, instead of every time the panel is activated or when leaving the options submenu.

Additionally, replace macros.index() with enumerate() while constructing the macro list.

### DIFF
--- a/panels/gcode_macros.py
+++ b/panels/gcode_macros.py
@@ -49,7 +49,8 @@ class Panel(ScreenPanel):
         self.labels["options_menu"].add(self.labels["options"])
 
     def activate(self):
-        self.reload_macros()
+        if not self.macros:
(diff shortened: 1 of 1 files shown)


alfrix:
Thank you for the PR but it has a big issue, this would break the visibility toggles

JP-Reitsma:
Good catch, thanks! I reworked the optimization so the macro list is marked dirty whenever a visibility option changes. reload_macros() now rebuilds the list only when necessary, which restores the visibility updates while preserving the optimization.

alfrix:
seems to be working thank you for helping improve KS

PR #1732, opened 2026-07-15
Render multi-color Spoolman spools as a gradient icon

4o66:
## Problem

Spoolman filaments store multi-color data in `multi_color_hexes` (and `multi_color_direction`) instead of `color_hex`. KlipperScreen only ever reads `color_hex`, so any multi-color spool renders as solid black in both the titlebar spoolman widget and the spool list/detail panel — `color_hex` is `null` for these filaments, and the code has no fallback.

I couldn't find an existing open issue for this (searched title/body for "spoolman", "spool", "color", "multicolor", "multi-color", "multi_color_hexes" — nothing matches), so this PR doesn't close anything specific.

## Fix

- `ks_includes/svg_gradient.py` (new): `filament_colors_from_dict()` resolves a spool's ordered color list, preferring `multi_color_hexes` and falling back to `color_hex`, then black. `apply_filament_gradient()` recolors a spool icon with a single smooth top-to-bottom gradient through those colors.
- Works generically across all five bundled themes' `spool.svg` files without any per-theme geometry: it clones whichever elements use `fill:var(--filament-color)` into an SVG `<mask>` (preserving each theme's exact filament-coil shape, whatever it looks like), then paints one gradient rect through that mask.
- Single-color spools are unaffected — same plain fill as before, byte-for-byte, no mask/gradient overhead.
- `panels/spoolman.py` and `panels/base_panel.py` updated to use the shared helper (spool list icon and titlebar icon respectively).

## Design note

This renders coaxial and longitudinal multi-color filaments identically (one gradient, first color nearest the spool's near face, shifting outward). I considered rendering these two cases differently but decided the added complexity wasn't worth it for a small titlebar/list icon — happy to revisit if maintainers feel otherwise.

## Testing

- `ruff check` / `ruff format` clean against the paths CI lints.
- Unit-level assertions on color resolution (multi vs. single vs. empty/whitespace vs. missing).
- Confirmed the single-color path is identical to the previous behavior.
- Rendered the actual shipped code against all 6 `spool.svg` files (default + 5 themes) to confirm no leftover `var(--filament-color)` placeholders and valid output.
- Tested live on two separate physical setups, both on the `z-bolt` theme: a Qidi X-Plus3 with KlipperScreen on a dedicated Pi, and a Trident 350 with KlipperScreen running on its Manta M8P mainboard and 5-inch HDMI screen (Siboor June AWD kit). Confirmed working on both.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Probably all versions contain changes regarding documentation, translation, fixe
 * network interface selector includes wired connections
 * add a shortcut from titlebar item to respective panels (temp/spoolman)
 * add a configurable screensaver delay for slow to wake displays
+* render multi-color spools (Spoolman multi_color_hexes) as a gradient in the spoolman icon, instead of black
 
 ## v0.4.7  (May 13, 2026)
 
--- a/ks_includes/svg_gradient.py
+++ b/ks_includes/svg_gradient.py
@@ -1,0 +1,107 @@
+import re
+
+# Matches any element using the `fill:var(--filament-color)` placeholder, regardless
+# of which theme's spool.svg it comes from or how that theme shapes the filament
+# coil (a plain rect, a curved wedge, one shape, two shapes, etc).
+FILAMENT_COLOR_RE = re.compile(r"<path[^>]*?fill:var\(--filament-color\)[^>]*?/>")
+SVG_TAG_RE = re.compile(r"(<svg\b[^>]*>)")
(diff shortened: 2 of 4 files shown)


thijstriemstra:
sounds great. love to see a screenshot of the new gradient icon(s)

4o66:
Here's a mockup of the spools, in three color "modes" across the stock themes. It should™ be theme agnostic, but I don't run anything but stock.

combined_final

alfrix:
Seems to work, i have no real complaints, thank you for helping improve KS

PR #1746, opened 2026-08-11
fix: ensure spoolman ID column does not truncate for values > 99

jmlw:
Issue: Spoolman cuts off IDs > 99

This PR fixes the rendering issue by creating and supplying a new renderer without ellipsize enabled and sets a minimum width to ensure the ID column is always rendered appropriately.

### Old:

ID column ellipsized not displaying IDs > 99

image

### New:

ID column with fixed width and no ellipsize mode configured

image


alfrix:
Hi, thank you for contributing to the project for the first time, the changes look good except the hardcoded min_width, but don't worry i'll remove that myself, if absolutely needed, the size should be in relative units according to the text done in css

PR #1749, opened 2026-08-14
Make notification popovers non-modal

### LINKED ISSUE #1748: [BUG] Success notifications block touchscreen input
### What happened?

When a Klipper macro is executed from KlipperScreen and completes successfully, KlipperScreen displays a green success notification.

While this notification is visible, it captures touchscreen input. Other buttons do not respond until the notification is tapped or automatically dismissed.

This was reproduced with the `AFC_BRUSH` macro.

### Version

KlipperScreen `v0.4.7`, from the official `KlipperScreen/KlipperScreen` repository.

The green success notification should remain visible, but the underlying KlipperScreen interface should remain usable. I should be able to press another button without first dismissing the notification.

1. Open KlipperScreen.
2. Open the macro menu.
3. Execute a macro that completes successfully.
4. While the green notification is visible, tap another button.
5. Observe that the other button does not respond.

### Additional information:

The notification is implemented using `Gtk.Popover`. GTK popovers are modal by default, and KlipperScreen does not currently override that behavior in `show_popup_message()`.

A possible fix would be to make informational success notifications non-modal while retaining modal behavior for warnings and errors.

### Log output

The attached log starts at KlipperScreen startup and includes the reproduction. The successful macro notifications appear at approximately `11:15:10`, `11:15:17`, and `11:15:23` AEST on 2026-08-14.

[KlipperScreen-v0.4.7-reproduction.log](https://github.com/user-attachments/files/31049951/KlipperScreen-v0.4.7-reproduction.log)

maddavo:
Fixes #1748

## What changed

All notification popovers are now non-modal, regardless of notification level. Success, warning, and error notifications retain their existing styling and timeout behaviour, but no longer capture touchscreen input.

## Why

KlipperScreen uses `Gtk.Popover` for notification messages. GTK popovers are modal by default, which means a visible notification can block interaction with the underlying interface until it is dismissed or times out.

Making all notification popovers non-modal gives the different notification levels consistent interaction behaviour and ensures notifications do not interfere with normal user control. Dedicated error dialogs remain modal where explicit user acknowledgement is required.

## Validation

- `python -m py_compile screen.py`
- `git diff --check`
- Reproduced on official KlipperScreen `v0.4.7`; see issue #1748 and its attached startup-to-reproduction log.

### DIFF
--- a/screen.py
+++ b/screen.py
@@ -505,7 +505,7 @@ class KlipperScreen(Gtk.ApplicationWindow):
             halign=Gtk.Align.CENTER,
             width_request=int(self.width * 0.9),
         )
-        popup.set_modal(level != 1)
+        popup.set_modal(False)
         popup.get_style_context().add_class("message_popup_popover")
         popup.add(msg)
         popup.popup()
--- a/screen.py
+++ b/screen.py
@@ -505,6 +505,7 @@ class KlipperScreen(Gtk.ApplicationWindow):
             halign=Gtk.Align.CENTER,
             width_request=int(self.width * 0.9),
         )
+        popup.set_modal(level != 1)
         popup.get_style_context().add_class("message_popup_popover")
         popup.add(msg)
         popup.popup()


maddavo (commented):
Updates nonmodality change from success notifications to all notifications.

alfrix:
Thank you for contributing and improving ks

PR #1755, opened 2026-08-21
Lazy-load KlippyRest

JP-Reitsma:
KlippyRest is now instantiated only when the REST API is actually needed.

This avoids creating the REST client during normal startup while preserving the existing REST API behavior.

Tested:

- KlipperScreen starts normally without initializing KlippyRest.
- REST API functionality triggers initialization when needed.
- git diff --check passes.

### DIFF
--- a/ks_includes/KlippyGtk.py
+++ b/ks_includes/KlippyGtk.py
@@ -198,7 +198,7 @@ class KlippyGtk:
             return
 
         def _load():
-            response = self.screen.restApi.get_thumbnail_stream(resource)
+            response = self.screen.get_rest_api().get_thumbnail_stream(resource)
             if response is False:
                 GLib.idle_add(callback, None)
                 return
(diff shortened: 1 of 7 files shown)


JP-Reitsma:
I tested this some more on a RK3308.

One thing I can confirm from testing is that lazy-loading KlippyRest moves a significant amount of initialization work to its first actual use. I measured the first KlippyRest initialization five times:

2961.6 ms
2974.9 ms
2979.8 ms
2977.9 ms
3029.6 ms

So the first use consistently incurs roughly a 3-second delay.

I don't want to make a stronger claim about the total startup-time difference, because my startup measurement did not include the import that happens before my startup marker.

However, moving ~3 seconds of initialization to first use is undesirable on this hardware. For example, when opening the shutdown screen during a printer problem, adding several seconds of latency before the REST API becomes available is a significant downside.

Based on this testing, I don't think this change is beneficial for this hardware, so I'm closing the PR rather than having it merged.

Thanks for the work on this.

PR #1756, opened 2026-08-21
fix: switch to configured VT before starting service

aaronmee:
When using KlipperScreen with Cage/Wayland, Cage starts on the currently active VT despite `TTYPath` being set to /dev/tty7.

This PR fixes the issue by switching to the configured VT before Cage initializes its DRM/KMS session, ensuring that KlipperScreen always starts on the intended VT.

My tests showed that this change is only required for the Cage/Wayland backend; the X11 backend is unaffected.

### DIFF
--- a/scripts/KlipperScreen.service
+++ b/scripts/KlipperScreen.service
@@ -20,6 +20,7 @@ SupplementaryGroups=klipperscreen
 WorkingDirectory=KS_DIR
 Environment="KS_XCLIENT=KS_ENV/bin/python KS_DIR/screen.py" BACKEND=KS_BACKEND
 ExecStart="KS_DIR/scripts/KlipperScreen-start.sh"
+ExecStartPost=+chvt 7
 
 # Log this user with utmp, letting it show up with commands 'w' and
 # 'who'. This is needed since we replace (a)getty.


alfrix:
lgtm thank you for contributing

JP-Reitsma:
PR #1756 adds ExecStartPost=+chvt 7 to the KlipperScreen service. On a QIDI Q2/RK3308 using Cage with the DRM backend, this fails with:

chvt: Couldn't get a file descriptor referring to the console.

This may also affect other KlipperScreen installations that don't use a traditional VT-based display setup. Could this be made conditional on a usable VT being available?

alfrix:
> PR #1756 adds ExecStartPost=+chvt 7 to the KlipperScreen service. On a QIDI Q2/RK3308 using Cage with the DRM backend, this fails with:
> 
> chvt: Couldn't get a file descriptor referring to the console.
> 
> This may also affect other KlipperScreen installations that don't use a traditional VT-based display setup. Could this be made conditional on a usable VT being available?

maybe we can use

```
ExecStartPost=+/bin/sh -c 'chvt 7 || true'
```

alfrix:
or we can simply target tty1 remove the switching

JP-Reitsma:
I think targeting a fixed VT would be cleaner than ignoring a failed chvt.

On the Q2 there is no usable VT at all, so chvt 7 || true would make the service start, but it would also silently hide a real VT-switching problem on systems where VT switching is expected to work.

Is there a reason to prefer tty1 over keeping the configured VT and only switching when a usable VT exists?

aaronmee:
Good catch with the missing VT error, I didn't think about systems withouth usable VT's.
This seems to be a general "issue" of KlipperScreen since it expects a working VT, thus you have to remove the line `ConditionPathExists=tty0` for it to work on the Q2 (per [community guide](https://github.com/qidi-community/q2-wiki/blob/main/content/klipperscreen-installation/README.md#service-configuration)).

I agree that ignoring a failed `chvt` isn't best practice as it introduces new possibilities of silent failures. IMO, the best general solution would be to make the VT requirements conditional, thus removing those lines on systems without usable VT's.

@JP-Reitsma as I don't have access to a non-VT system, could you tell me if your system lists any `tty*` under `/dev`?

JP-Reitsma:
@aaronmee There are only:

/dev/tty (process terminal)
/dev/ttyFIQ0 (serial console)
/dev/ttyS4 (serial UART)

No /dev/tty0 through /dev/tty7...

(I'm using the mainline setup with an addded CB1 handling Klipper, while the RK3308 is only used for the display/KS. So this isn't the stock Q2 architecture from the community guide. The RK3308 nevertheless has the same lack of virtual terminals.)

aaronmee:
@JP-Reitsma below are some ideas I aquired. I'm unsure whether or not they will do the trick to detect a (non) funtional VT environment.
For me (working VT), both return interactive/usable.

Determine whether bash is running interactively
```bash
if [ -z "$PS1" ]; then
    echo This shell is not interactive
else
    echo This shell is interactive
fi
```

Test whether /dev/tty is writable
```bash
if sh -c ": >/dev/tty" > /dev/null 2>&1; then
    echo /dev/tty is available and usable
else
    echo /dev/tty is not available
fi
```

JP-Reitsma:
@aaronmee I think we can make this conditional without hiding real `chvt` failures:

INI

```
ExecStartPost=+/bin/sh -c '[ -e /dev/tty0 ] && [ -e /dev/tty7 ] && chvt 7'
```

On systems without Linux VTs, `chvt` is skipped. On VT-based systems, the existing behaviour remains and a real `chvt` failure is still reported.

aaronmee:
@JP-Reitsma tell me your opinion on #1764, which should ideally resolve this discussion.

PR #1757, opened 2026-08-23
Remove klippyrest

JP-Reitsma:
## Background

Testing the lazy-loading PR showed that the first get_rest_api() call took about 3 seconds on the RK3308 (5 measurements: 2961.6–3029.6 ms, average 2985 ms).

This raised the question whether `KlippyRest` is still needed at all, given that KlipperScreen already maintains a persistent WebSocket connection to Moonraker.

I reviewed all six current REST API call sites:

1. **`machine/system_info`**  
   Already retrieved during initialization through the existing WebSocket (`machine.system_info`) and stored in `printer.system_info`. The REST call was only a fallback and has been removed.

2. **`server/history/job`**  
   A WebSocket equivalent (`server.history.get_job`) already exists and is used elsewhere in KlipperScreen. The remaining REST call was converted to use the existing WebSocket API.

3. **`server/database/item` / Mainsail LED presets**  
   The existing code relied on the obsolete `mainsail/miscellaneous.entries` database structure. Current Mainsail versions use a different LED/presets implementation, so this obsolete lookup has been removed. Support for the current Mainsail LED/effects functionality can be addressed separately.

4. **`camera.endpoint`**  
   No REST request was made here. `KlippyRest` was only used to construct the Moonraker base URL for relative camera URLs. This is now handled directly from the existing connection configuration.

5. **`thumbnail_stream`**  
   This is the only remaining genuine HTTP use case. G-code thumbnails are downloaded from Moonraker's HTTP file API and are already loaded asynchronously when needed. The HTTP request is now performed directly with `requests`, with `requests` itself imported only when a thumbnail is actually loaded.

6. **`shutdown.endpoint`**  
   `KlippyRest` was only used to log its endpoint. The actual reboot/shutdown operations already use the existing WebSocket (`machine.reboot` / `machine.shutdown`), so the REST dependency was removed.

## Result

`KlippyRest` was only required as a general-purpose wrapper around a small number of legacy or duplicated REST calls. After removing the obsolete uses and moving the remaining thumbnail download to a direct HTTP request, the `KlippyRest` class itself can be removed entirely.

The resulting code has fewer lines overall and retains HTTP access where it is actually required, without maintaining a separate REST client.

## Testing

Tested on a RK3308.

| Version | Startup times | Average |
| --- | --- | ---: |
| `master` | 14.874, 14.854, 14.846, 14.950, 14.649 s | **14.835 s** |
| `remove-klippyrest` | 11.530, 11.780, 11.683, 11.388, 11.590 s | **11.594 s** |

This is a reduction of **3.241 seconds (21.8%)** in the measured startup time.

The `requests` import is intentionally performed on demand when a thumbnail is requested. The G-code list itself is available immediately, while thumbnails are loaded asynchronously afterwards.

- System information continues to work.
- Job history / Last Duration continues to work.
- Camera streaming continues to work.
- G-code thumbnails continue to load correctly.
- The code passes `py_compile`.
- No remaining `KlippyRest`, `restApi` or `get_rest_api()` references remain.


alfrix:
the difference might be exaggerated by the potato cpu, but you are right, i've moved most of the calls to the websocket and this was left to do, thank you

PR #1758, opened 2026-08-23
Skip Cage installation when already installed

JP-Reitsma:
When Wayland/Cage is selected, the installer currently attempts to install Cage even when it is already available on the system.

This change checks whether the `cage` executable is already present and skips the installation when it is.

### DIFF
--- a/scripts/KlipperScreen-install.sh
+++ b/scripts/KlipperScreen-install.sh
@@ -44,6 +44,12 @@ install_graphical_backend()
       read -r -e -p "Backend X11 or Wayland (cage)? [X/w]" BACKEND
     fi
     if [[ "$BACKEND" =~ ^[wW]$ ]]; then
+        if command -v cage >/dev/null 2>&1; then
+            echo_ok "Cage already installed"
+            BACKEND="W"
+            break
+        fi
+
         echo_text "Installing Cage"
         if sudo apt install -y $CAGE; then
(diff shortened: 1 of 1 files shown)


alfrix:
Not needed, apt already skips already installed packages

JP-Reitsma:
My Q2 runs Debian 11 (Bullseye), using the standard Bullseye repositories plus the Mainsail repository. The cage package is not available on this potato, so the normal apt install cage path cannot provide Cage.

It therefore uses a locally built Cage 0.1.2 / wlroots 0.11 stack under /usr/local. The command -v cage check therefore allows the installer to detect an existing Cage installation even when it is not managed by apt/dpkg.

alfrix:
but you are not checking seatd, which is also needed

perhaps instead of this system we could have the installer to ask what to install

Cage
Weston
X11
Custom / Skip

and modify the start script to start the first one available

JP-Reitsma:
I agree that would be a cleaner approach than this check. 
In my Q2 setup, Cage is a custom local build under /usr/local, while seatd is installed normally as a Debian package and runs as a systemd service.My Q2 setup was what exposed the issue, so I'm happy to leave the PR as-is for reference and let you decide whether to replace it with the more general approach.

JP-Reitsma:
Thanks @alfrix for working on #1772 . Looks good, I'm closing this PR in favor of that approach.

